PR #83, opened 2015-01-07
Death queue timeout set does not match the comments.

pjhades:
As the comments say, the death queue timeout should only be set when a connection is keep-alive, AND its related coroutine needs to be resumed.


pjhades:
Sorry I made a mistake here. :(
The original code is correct. I have deleted the change of the bit operator.

PR #88, opened 2015-03-03
Removes a unseless field in a struct of lwan-http-authorize.c

RaphaelJ:
The struct is not exposed to other compilation units, and the field is not used
in the compilation unit at all. Thus the field seems useless to me.


lpereira:
The `base` field isn't useless. It's embedded in the structure as the first member on purpose, since the caching layer will store the cache entry metadata there (refcount, flags, etc).

PR #90, opened 2015-03-15
Minor changes

damag:
Valgrind error fix and minor comment cleanup.

### DIFF
--- a/common/lwan-request.c
+++ b/common/lwan-request.c
@@ -566,7 +566,9 @@ static lwan_read_finalizer_t read_request_finalizer(size_t total_read,
     char *terminator = memmem(helper->buffer->value, helper->buffer->len, "\n\r\n", 3);
     if (LIKELY(terminator)) {
         char *method = terminator + 3;
-        if (get_http_method(method) && terminator != helper->buffer->value) {
+        if ((size_t)(method - helper->buffer->value) < helper->buffer->len &&
+                get_http_method(method) &&
+                terminator != helper->buffer->value) {
             helper->next_request = method;
             return FINALIZER_DONE_PIPELINED;
         }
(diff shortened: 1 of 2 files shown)


damag:
I just realized, should the buffer actually check past the method pointer since the get_http_method() will actually read 4 bytes starting at that location? The current commit resolves the issue when testing with curl or a browser, but some weird requests could still trigger the valgrind error if (buffer + len - method) is less than 4.

lpereira:
I've been thinking, and this part of the code will need to go away eventually (or change).

The `parse_headers()` function should find the end of the request already; I haven't tested but it seems it's not really doing that; it's parsing all the pipelined requests over and over (amortized `O(n²)`).

If it gets fixed, then there's no need to call `memmem()` to find the terminator, and that `memmove()` will only need to happen if the headers are being parsed before a `\r\n\r\n` is found (although this is tricky as the request will essentially have to be reparsed to patch the pointers to the buffer).

damag:
Yea, the code in `parse_headers()` seems to assume that the entire buffer is a single request:

        if (p >= buffer_end)            /* reached the end of header blocks */ \
          goto end; \

and in the `did_not_match` section it breaks out of the loop once we reach reach the '\0' at the end of the buffer instead of breaking out after this request's headers have been parsed.

Anyways, I guess there's no point in adding code churn to something that will get replaced. Feel free to close this pull request (although that first commit with the minor comment change may still be helpful). :)

lpereira:
With 63f1e58, the error reported should be gone; things were simplified and are much closer to what they were before pipeline has been introduced (no mode switch, etc). I'm still not happy with the parser, though, but given that it should be revisited if HTTP/2 is to be implemented, I'm not worrying about it right now.

I've cherry-picked the other commit from this PR and pushed everything to master.

PR #91, opened 2015-03-25
Set specific CMAKE_C_FLAGS_* for each build type

damag:
This avoids calling the compiler with multiple conflicting optimization
flags, since both CMAKE_C_FLAGS_RELEASE and CMAKE_C_FLAGS are used when
doing a release build.

Also explicitly set "Debug" as the default CMAKE_BUILD_TYPE. Otherwise
CMake seems to default to a release build.


lpereira:
Good one! One minor nit, though: in order to not duplicate flags for both `CMAKE_C_FLAGS_RELEASE` and `CMAKE_C_FLAGS_RELWITHDEBINFO`, it might be a good idea to copy, just once, the value of `CMAKE_C_FLAGS_RELEASE` to `CMAKE_C_FLAGS_RELWITHDEBINFO` right before including other directories

damag:
Yea I agree it is error-prone to have the same flags in multiple places. I'm not sure what you mean by copying the value of `CMAKE_C_FLAGS_RELEASE` into `CMAKE_C_FLAGS_RELWITHDEBINFO`. Here are what the initial values appear to be before setting anything:
```
CMAKE_C_FLAGS_DEBUG = "-g"
CMAKE_C_FLAGS_RELEASE = "-O3 -DNDEBUG"
CMAKE_C_FLAGS_RELWITHDEBINFO = "-O2 -g -DNDEBUG"
CMAKE_C_FLAGS_MINSIZEREL = "-Os -DNDEBUG"
```
Are you saying we should set our own values for each rather than appending to CMake's defaults? I think we should set a temporary variable to append to the various release build types, like:

```
set(C_FLAGS_REL "-mtune=native -flto")
...
set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} ${C_FLAGS_REL}")
set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} ${C_FLAGS_REL}")
set(CMAKE_C_FLAGS_MINSIZEREL "${CMAKE_C_FLAGS_MINSIZEREL} ${C_FLAGS_REL}")
```
Also, `CMAKE_C_FLAGS` seems to be initialized by the `CFLAGS` environment variable, not matter which built type is selected. Do we want to set that rather than append to it since that might also contain conflicting flags?

On a side note, we should probably also support MinSizeRel for completeness.

damag:
I've updated the commit to accumulate the release flags into another variable (`C_FLAGS_REL`) like I mentioned above. Let me know if this is what you had in mind.

lpereira:
Yeah, I think this looks cleaner without all the duplicated lines. Merging, thank you!

PR #94, opened 2015-04-07
Add 'request_id' to request struct and populate with uuid

malthe:
The request id is a universally unique identifier (specifically, a UUID version 1) that can be attached to logging output as well as added to the response, for instance in the form of a 'X-Request-Id' header.

### DIFF
--- a/common/hash.c
+++ b/common/hash.c
@@ -84,7 +84,8 @@ void hash_random(void *p, unsigned short size)
             return;
     }
 
-	assert(read(fd, p, size) == size);
+    size -= read(fd, p, size);
+    assert(!size);
     close(fd);
 }
 
--- a/common/hash.c
+++ b/common/hash.c
@@ -21,6 +21,7 @@
 #include "hash.h"
 #include "murmur3.h"
 
+#include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <stdbool.h>
@@ -37,6 +38,7 @@ enum {
 	steps = 64,
 	default_odd_constant = 0x27d4eb2d
 };
+
 static unsigned odd_constant = default_odd_constant;
 
 struct hash_entry {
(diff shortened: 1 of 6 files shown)


common/hash.c
```diff
@@ -63,26 +65,27 @@ struct hash {
...
-	if (read(fd, &odd_constant, sizeof(odd_constant)) != sizeof(odd_constant))
-		odd_constant = default_odd_constant;
-	close(fd);
+        if (fd < 0)
+            return;
+    }
 
-oddify_constant:
-	odd_constant |= 1;
+	assert(read(fd, p, size) == size);
```
lpereira: `assert()` will expand to nothing in release mode.

common/hash.h
```diff
@@ -15,6 +15,7 @@ struct hash *hash_int_new(void (*free_key)(void *value),
 			void (*free_value)(void *value));
 struct hash *hash_str_new(void (*free_key)(void *value),
 			void (*free_value)(void *value));
+void hash_random(void *p, unsigned short size);
```
lpereira: Why is this being exposed? The random in the hash is only used to ensure the bucket each item ends up being in is unpredictable; it's an implementation detail.

common/lwan-thread.c
```diff
@@ -309,6 +380,7 @@ thread_io_loop(void *data)
     const int max_events = min((int)t->lwan->thread.max_fd, 1024);
 
     lwan_status_debug("Starting IO loop on thread #%d", t->id + 1);
+    hash_random(&t->node, 8);
```
lpereira: `sizeof(&t->node)` is always 8?

common/hash.c
malthe: I know. I don't expect it to be false in production.

lpereira:
This looks like it adds a considerable amount of overhead (at least 1 syscall, but may be as high as 3 if `getentropy()` is not available, plus all the math, string formatting, etc) for not much gain. Is there any benefit I can't see that would outweigh that?

common/hash.c
lpereira: The read won't even happen as the expression will not even be generated. assert() should only be used to ensure invariants are respected.

malthe: Oh dang. Good catch.

common/hash.h
malthe: I need to move this to a separate module `hash.c` because I use it in two different places now.

common/lwan-thread.c
malthe: It's at least 8 and I only ever use 8 bytes from it (converted to big-endian first).

malthe:
In my informal benchmarks there is only a negligible overhead – where do you see the syscall? There is a call to `getrandom()` but that's only done once per thread initialization.

lpereira:
This has been implemented slightly differently (without UUIDs, just a 64-bit number). I apologize for the immense delay.

PR #97, opened 2015-04-25
Make it easier to use lwan with C++ code.

byzhang:
Enclose lwan.h by extern "C" for CXX code, and add a CXX friendly declaration of lwan_request_get_remote_address and lwan_format_rfc_time, as static array parameter is C99 only.


lpereira:
Merged. Thank you!

PR #98, opened 2015-04-26
Install liblwan-common.a and header files.

byzhang:


lpereira:
Nice!

I'd just add a [`.pc` file](https://wiki.freedesktop.org/www/Software/pkg-config/) in order to ensure the program that uses the library also uses the same linking options that would be used if built alongside the Lwan build system (setting `Libs.private` and whatnot).

byzhang:
Didn't use pkg-config much before. Maybe you can merge this pull request and have a follow up commit? Or you may just make a new commit from scratch. Anyway, lwan is handy for an embedding http server :)

lpereira:
I'll merge this commit and make the pkg-config stuff.

lpereira:
Merged, thanks!

PR #100, opened 2015-05-01
Make it easier to use as a lib

byzhang:
moving json from techempower to common;
making some char* const.

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -69,6 +69,6 @@ set(ADDITIONAL_LIBRARIES ${ADDITIONAL_LIBRARIES} PARENT_SCOPE)
 INSTALL(TARGETS lwan-common
   DESTINATION "lib"
 )
-INSTALL(FILES lwan.h lwan-coro.h lwan-trie.h lwan-status.h strbuf.h hash.h lwan-template.h
+INSTALL(FILES lwan.h lwan-coro.h lwan-trie.h lwan-status.h strbuf.h hash.h lwan-template.h lwan-serve-files.h
   DESTINATION "include/lwan"
 )
--- a/common/lwan.h
+++ b/common/lwan.h
@@ -217,7 +217,7 @@ struct lwan_url_map_t_ {
     lwan_http_status_t (*handler)(lwan_request_t *request, lwan_response_t *response, void *data);
     void *data;
 
-    char *prefix;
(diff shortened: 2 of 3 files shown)


common/lwan.h
```diff
@@ -304,6 +304,7 @@ int lwan_connection_get_fd(lwan_connection_t *conn)
     __attribute__((pure)) __attribute__((warn_unused_result));
 
 #if defined (__cplusplus)
+#include "lwan-serve-files.h"
```
lpereira: I wouldn't include this file here if using C++: let the user include both `lwan.h` and `lwan-serve-files.h`.

common/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ set(SOURCES
 	base64.c
 	hash.c
 	int-to-str.c
+  json.c
```
lpereira: Indentation seems off here.

common/lwan.h
byzhang: Fixed.

common/CMakeLists.txt
byzhang: Fixed.

techempower/CMakeLists.txt
```diff
@@ -21,7 +21,6 @@ if (MYSQL_INCLUDE_DIR AND SQLITE_FOUND)
 
 		add_executable(techempower
 			techempower.c
-			json.c
```
byzhang: Json decoding is pretty helpful to parse the POST data. And lwan is a high performance engine for REST services which also needs both JSON encoding and decoding.

lpereira:
Sorry for taking so long to reply, I've been pretty busy with the template parser rewrite.

I'd make JSON as another library that your program can link to, at least when I've fully reviewed that code. I've picked that library for the sole purpose of making it to the Techempower benchmark. It's fast and has a sane API, but I need to understand all of it before it can go into the core. That's easy to do with CMake, though.

Also, please make different changes in different commits. Making a lot of unrelated changes in one single commit breaks things like git bisect, which are pretty useful to find regressions.

It's OK to force-push to the branch you've used to make the pull request, so using 'git rebase -i'  to squash/fixup/reorder commits is perfectly fine.

Cheers!

byzhang:
updated.

lpereira:
Merged, thank you!

PR #105, opened 2015-06-11
Remove off by one error in internal_printf

dallbee:
when called here, save_str is expecting a size, but is given a length.

### DIFF
--- a/common/strbuf.c
+++ b/common/strbuf.c
@@ -211,7 +211,7 @@ internal_printf(strbuf_t *s1, bool (*save_str)(strbuf_t *, const char *, size_t)
     if (UNLIKELY((len = vasprintf(&s2, fmt, values)) < 0))
         return false;
 
-    bool success = save_str(s1, s2, (size_t)len);
+    bool success = save_str(s1, s2, (size_t)len + 1);
     free(s2);
 
     return success;


lpereira:
I'm curious here: were you able to trigger a bug with the current behavior?

lpereira:
I'm curious here: were you able to trigger a bug with the current behavior?

dallbee:
Yep! That's how I found it.

I had a page handler doing the following: 
```c
strbuf_t *filename = strbuf_new();
strbuf_printf(filename, "%s.md", req->url.value);
// do stuff with filename (nothing mutating it though)
strbuf_free(filename);
```
First page load: string is as expected.
Second (and additional) page load: string has an additional, random character appended to it.

When run with GDB, the problem vanishes, because GDB will use a CRT debug heap, memsetting the data to 0 each time it is freed. Since the string is looking for a null terminating character, everything is happy.

Valgrind had a similar issue of hiding the bug, and so my friend and I had been calling this thing a Heisenbug for the last week (it's taken me a while to trace this one down).

lpereira:
Ouch.
Merged, thanks!

lpereira:
Hm, it seems that the tests are failing now. The bug is likely somewhere else. I'm reverting the patch.

Issue #117, opened 2015-10-18
PROXY protocol


malthe:
The [Hitch TLS proxy](https://hitch-tls.org/) speaks through the simple [PROXY protocol](http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt) to a backend server, e.g.

    PROXY TCP4 192.168.0.1 192.168.0.11 56324 443\r\n
    GET / HTTP/1.1\r\n
    Host: 192.168.0.11\r\n
    \r\n

Supporting this protocol makes #12 less important, perhaps obviating it entirely.

lpereira:
This issue can now be closed.

PR #118, opened 2015-10-18
Add support for proxy protocol

malthe:
Note that this adds a ``struct lwan_proxy_t``:
```c
l->proxies = calloc(max_open_files, sizeof(lwan_proxy_t));
``` 
This is a waste of memory when PROXY protocol isn't used. It might be an idea to have a flag that controls whether this is always allocated, lazily allocated or never allocated.

It might be better to add this to the coro's stack?

Fixes https://github.com/lpereira/lwan/issues/117.

### DIFF
--- a/common/lwan-request.c
+++ b/common/lwan-request.c
@@ -100,6 +100,8 @@ get_http_method(const char *buffer)
 static lwan_request_flags_t
 parse_proxy_protocol(lwan_request_t *request, char **buffer, int version)
 {
+    if (!request->conn->thread->lwan->config.proxy_protocol) return 0;
+
     union header_t_ {
         struct {
             char line[108];
--- a/common/lwan-request.c
+++ b/common/lwan-request.c
@@ -70,9 +70,12 @@ get_http_method(const char *buffer)
 {
     /* Note: keep in sync in identify_http_method() */
     enum {
-        HTTP_STR_GET  = MULTICHAR_CONSTANT('G','E','T',' '),
-        HTTP_STR_HEAD = MULTICHAR_CONSTANT('H','E','A','D'),
-        HTTP_STR_POST = MULTICHAR_CONSTANT('P','O','S','T')
+        HTTP_STR_GET     = MULTICHAR_CONSTANT('G','E','T',' '),
+        HTTP_STR_HEAD    = MULTICHAR_CONSTANT('H','E','A','D'),
+        HTTP_STR_POST    = MULTICHAR_CONSTANT('P','O','S','T'),
+        HTTP_STR_PROXY1  = MULTICHAR_CONSTANT('P','R','O','X'),
+        HTTP_STR_PROXY21 = MULTICHAR_CONSTANT('\x00','\x0D','\x0A','\x51'),
+        HTTP_STR_PROXY22 = MULTICHAR_CONSTANT('\x55','\x49','\x54','\x0A')
     };
 
     STRING_SWITCH(buffer) {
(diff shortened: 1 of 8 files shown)


common/lwan.c
```diff
@@ -466,8 +466,14 @@ static void
 allocate_connections(lwan_t *l, size_t max_open_files)
 {
     l->conns = calloc(max_open_files, sizeof(lwan_connection_t));
-    if (!l->conns)
-        lwan_status_critical_perror("calloc");
+    if (!l->conns) goto err;
+
+    l->proxies = calloc(max_open_files, sizeof(lwan_proxy_t));
```
lpereira: Maybe each `lwan_request_t` should have a `lwan_proxy_t`, so there's no need to malloc this here (which might lead to false sharing, which is why the connection structure is 64-byte long on x86-64).

common/lwan-request.c
```diff
@@ -82,23 +85,260 @@ get_http_method(const char *buffer)
...
+        if ((string_as_int32(buffer + 4) == HTTP_STR_PROXY22) &&
+            (*(buffer + 8) >> 4 == 2))
+            return REQUEST_METHOD_PROXY2;
     }
 
     return 0;
 }
 
+static lwan_request_flags_t
+parse_proxy_protocol(lwan_request_t *request, char **buffer, int version)
```
lpereira: Maybe this function can be split in two, one for each version, and have `identify_http_method()` call them directly? This should avoid the `if (version == V)` branches.

common/lwan-request.c
```diff
@@ -82,23 +85,260 @@ get_http_method(const char *buffer)
...
+    return get_http_method(*buffer);
+}
+
 static ALWAYS_INLINE char *
 identify_http_method(lwan_request_t *request, char *buffer)
 {
+    char *path;
     lwan_request_flags_t flags = get_http_method(buffer);
+
+    if (flags == REQUEST_METHOD_PROXY1)
```
lpereira: Since the proxy method allows one to essentially spoof the source IP address, it might be a good idea to make this an option and make it off by default. (So only people that require the proxy feature will enable it.)

common/lwan-request.c
```diff
@@ -82,23 +85,260 @@ get_http_method(const char *buffer)
...
+
+                    if (inet_pton(AF_INET, src_addr, &from->sin_addr) != 1) {
+                        return 0;
+                    }
+
+                    if (inet_pton(AF_INET, dst_addr, &to->sin_addr) != 1) {
+                        return 0;
+                    }
+
+                    porttmp = strtol(src_port, NULL, 10);
```
lpereira: Maybe use `strtoul()` here (as it won't look for negative numbers).

Setting the `endptr` is also a nice idea: check if it is one byte off from dst_port, just as a sanity check.

Might be a good idea to zero out `errno` before calling `strtoul()` as well. (See the man page, that's the recommended way.)

tools/testsuite.py
```diff
@@ -479,6 +479,31 @@ def test_cache_mmaps_once_even_after_timeout(self):
     requests.get('http://127.0.0.1:8080/100.html')
     self.assertEqual(self.count_mmaps('/100.html'), 1)
 
+class TestProxyProtocolRequests(SocketTest):
```
lpereira: Are proxied pipelined requests working? What about keep-alive requests? And other methods, such as POST?

common/lwan-request.c
```diff
@@ -882,14 +1122,26 @@ const char *
...
-    if (UNLIKELY(getpeername(request->fd, (struct sockaddr *)&sock_addr, &sock_len) < 0))
-        return NULL;
+    struct sockaddr_storage __sock_addr;
+    struct sockaddr_storage *sock_addr = &__sock_addr;
 
-    if (sock_addr.ss_family == AF_INET)
-        return inet_ntop(AF_INET, &((struct sockaddr_in *)&sock_addr)->sin_addr,
+    if (request->conn->flags & CONN_PROXIED) {
+        sock_addr = (struct sockaddr_storage *)
+            &request->conn->thread->lwan->proxies[request->fd].from;
```
lpereira: Yeah, definitely having an embedded proxy struct in the request struct will help cut down the indirection.

lpereira:
I'll take a more careful look this week, but so far, the patches are looking good. Thanks!

common/lwan-request.c
```diff
@@ -82,23 +85,262 @@ get_http_method(const char *buffer)
...
-        [REQUEST_METHOD_POST] = sizeof("POST ") - 1,
+        [REQUEST_METHOD_GET] = sizeof("GET") - 1,
+        [REQUEST_METHOD_HEAD] = sizeof("HEAD") - 1,
+        [REQUEST_METHOD_POST] = sizeof("POST") - 1,
     };
-    request->flags |= flags;
-    return buffer + sizes[flags];
+
+    if (flags) {
+        path = buffer + sizes[flags] + 1;
```
lpereira: Is this really necessary? The `[0] = 0` thing was there specifically to avoid this branch here. (That's why the code that  calls `identify_http_method()` actually checks if the buffer didn't change instead of checking for `NULL` as it was previously.)

common/lwan-thread.c
```diff
@@ -164,7 +164,7 @@ process_request_coro(coro_t *coro)
 
         next_request = lwan_process_request(lwan, &request, &buffer, next_request);
         if (!next_request)
-            break;
+            coro_run_deferred(coro);
```
lpereira: Are this and the next changes really necessary to make proxy work?

Deferred code is batched and executed every 16 resets (or a coroutine is destroyed, whichever comes first). This effectively bypasses that, possibly leading to resource exhaustion. Also, as commented in another PR, this is a dangerous API to expose: it can lead to race conditions and hard to track bugs, such as use after free, and whatnot.

common/lwan-request.c
lpereira: Splitting the function in two will also reduce the indentation level, making the function easier to read.

common/lwan-request.c
```diff
@@ -82,23 +85,262 @@ get_http_method(const char *buffer)
         return REQUEST_METHOD_HEAD;
     case HTTP_STR_POST:
         return REQUEST_METHOD_POST;
+    case HTTP_STR_PROXY1:
+        if (buffer[4] != 'Y') break;
```
lpereira: Most of the parsing code assumes that if prefixes matches, the rest is probably fine. So I'd say that it's OK to not check if `buffer[4]` is really `Y`, and just check if the rest of the things that must be absolutely checked are in the expected place.

lpereira:
Merged with changes, thank you!

PR #119, opened 2015-10-20
We don't need to break out of coro

malthe:
Also, this means the coro is always running when keeping the connection alive.

### DIFF
--- a/common/lwan-coro.c
+++ b/common/lwan-coro.c
@@ -147,7 +147,7 @@ coro_entry_point(coro_t *coro, coro_function_t func)
     coro_yield(coro, return_value);
 }
 
-static void
+void
 coro_run_deferred(coro_t *coro)
 {
     for (coro_defer_t *defer = coro->defer; defer;) {
--- a/common/lwan-coro.c
+++ b/common/lwan-coro.c
@@ -32,8 +32,6 @@
 #include <valgrind/valgrind.h>
 #endif
 
-#define CORO_RESET_THRESHOLD	16
-
 #define CORO_STACK_MIN		((3 * (PTHREAD_STACK_MIN)) / 2)
 
 static_assert(DEFAULT_BUFFER_SIZE < (CORO_STACK_MIN + PTHREAD_STACK_MIN),
@@ -44,6 +42,7 @@ typedef struct coro_defer_t_	coro_defer_t;
 struct coro_defer_t_ {
     coro_defer_t *next;
     void (*func)();
+    bool sticky;
     void *data1;
     void *data2;
 };
@@ -60,7 +59,6 @@ struct coro_t_ {
     coro_defer_t *defer;
     void *data;
 
-    unsigned char reset_count;
     bool ended;
 };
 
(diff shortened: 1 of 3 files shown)


lpereira:
The coroutines must be reset every now and then so that the `coro_defer` can be executed. For instance, when serving files, their file descriptors will only be closed whenever those deferred calls are executed. Some memory is never freed if these deferred stuff is not executed.

As a side note: It is possible to expose `coro_run_deferred()`, but that's a pretty dangerous API. In fact, I've just recently reverted a commit that did just that. There could be some sort of generational GC, though, where each reset would mark a generation, and have an exposed garbage collector, that would only execute the deferred code for obsolete generations. Not sure if that's worth the trouble, though.

malthe:
@lpereira in 32563c6 I do expose ``coro_run_deferred()``. I do see how it's a pretty dangerous API and it would be nice to avoid exposing it.

That said, compared to the situation before I don't think there's any difference. The coro will end if there are no pipelined requests and it will be reset when the next request comes in on the same file descriptor.

lpereira:
Not true for keep-alive connections, which are more common than pipelined requests in practice.

I've executed Lwan under strace and hit it with weighttp while it was serving a file with `sendfile()`, and I could see the file descriptors piling up. Without the patch, files are closed every 16 requests on the same file descriptor, as expected.

What are you trying to accomplish with this patch? Did you observe an undesirable behavior?

malthe:
Unless I'm missing something, the previous behavior was to exit the coro when ``next_request`` was null. Which it would be if there was no request in the pipeline.

That is, even with keep-alive enabled, the coro would exit and be reset between each request.

lpereira:
Yes, you're correct about the current behavior. It's the desired behavior. Unless you're seeing something I'm not, of course, in which case I'd like to understand why it's the wrong behavior so that we can fix it.

In any case, the proposed changes won't work:

* The reason `coro_run_deferred()` is not exposed is that calling while the coroutine is live can lead to crashes and other unexpected and often hard-to-debug behavior.
  * For instance, with this PR, the `strbuf` will be destroyed after the pipeline empties, and then will proceed to use deallocated memory.
* The coroutine not ending cleanly will change the behavior of `resume_coro_if_needed()`, possibly not  calling `epoll_ctl()` to toggle the write/read readiness flag.
* This change does not play nice with the batched `coro_defer` feature, which only runs the deferred code after a set number of resets.

malthe:
I guess I don't understand why a pipelined request is fundamentally different than a non-pipelined request. In theory, pipelined requests could be processed in parallel in some cases, but we don't do this here. It's still FIFO. Why break out of the coro in the general case and stay in it for the pipelined case?

lpereira:
The main difference is that, when attending to pipelined requests, there's usually no need to call `read()` on the socket file descriptor to get the next request.

The request buffer is reused (rather, `memmove()`d to the beginning of the buffer, see `read_from_request_socket()` in `lwan-request.c`), until it depletes or the request is incomplete. When that happens, `read_from_request()` will toggle an `epoll` flag to ensure the coroutine can be awoken whenever its associated file descriptor can be read again. The request will be read as if there was a short read the first time `read()` was called. The `epoll` flag will be reset, and the loop will continue to consume requests from the request buffer.

It's still FIFO, sure, but just avoiding the cost of the `read()` syscall (and it's associated `epoll` flag changes, which are also syscalls) is enough to ensure that pipelined requests are actually faster than non-pipelined requests.

There's a problem here, in that while pipelined requests are being handled, the `coro_defer` stuff is not executed until the whole pipeline is over. That's a thing that must be fixed but I'm not sure yet how to do this. That part of the code is pretty tricky and every time I touch that without thinking for a while, I have to fix a bunch of very weird bugs.

malthe:
I suppose that after ``CORO_RESET_THRESHOLD`` requests the right thing to do is to break out of the coro to prompt ``coro_defer()`` to run next.

lpereira:
Yes, limiting the number of pipelined requests (or number of requests per connection) is currently the easiest way to fix the problem. There are some things to consider, though. For instance, this might leave one (or a few) pipelined requests in mid-air, which is going to break things; the buffer (currently being allocated on the stack) would have to survive the coroutine being reset.

malthe:
In 5f0b136 I have added a ``sticky`` flag on a deferred which means it won't be run except when the coro terminates.

This flag is then applied to the response buffer deferred dealloc.

lpereira:
Merged with changes, thanks!

PR #120, opened 2015-11-08
Move switcher to thread struct to allow coro injection from outside

malthe:
I use this to co-opt the event loop with a redis connection per thread.

In my setup I connect to redis, clone the file descriptor and set up a reader and writer `coro_t` that both run as `lwan_connection_t` connections.

To inject the coros into the connection pool I use the same shared `switcher` (per thread).

### DIFF
--- a/common/lwan-thread.c
+++ b/common/lwan-thread.c
@@ -327,7 +327,7 @@ thread_io_loop(void *data)
     const lwan_t *lwan = t->lwan;
     lwan_connection_t *conns = lwan->conns;
     struct epoll_event *events;
-    coro_switcher_t switcher;
+    coro_switcher_t *switcher = &t->switcher;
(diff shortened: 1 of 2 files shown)


dyu:
@malthe is that lwan+redis integration in a public repo somewhere?
Its a very good lwan usecase/example that would benefit others in how to integrate faster.

malthe:
@dyu no it's not in a working state yet – it's unrealistic to implement it correctly before we a per-thread command queue is implemented, "including things like 'add file descriptor to epoll', 'make coroutine resumable', etc.".

That will make it much easier to co-opt lwan's io-loop with a redis per-thread worker.

dyu:
@malthe I see.  The scheduler needs some tweaking indeed to be flexible to other events.

PR #124, opened 2015-12-29
resolve FIXME comment in get_config_path()

LAC1213:
Load the config file using `argv[0]` or environment variable `LWAN_CONF`


lpereira:
Thanks for the PR! I do have a question, though. Commented in line.

LAC1213:
To be honest I don't know an urgent reason to switch to `argv[0]`, I just saw the comment and fixed it. But I think `argv[0]` is more intuitive than `readlink()`.  Some users might want to indicate their own location so I added support for the `LWAN_CONF` environment variable.

PR #129, opened 2016-03-04
Add FreeBSD package install names & instructions

koobs:
Add FreeBSD binary package names and install instructions using pkg(8)

While I'm here, rename "distributions" to "operating systems"


lpereira:
Merged, thanks!

PR #131, opened 2016-03-05
Fix is_hex_digit to properly recognize number digits.

matt-42:


lpereira:
Good catch! Merged, thanks.

PR #145, opened 2016-05-25
Build on Mac OSX

LampmanYao:
- Implemente pthread_barrier_t related APIs on mac osx
- Change some options when linking. Because of those options of the linker on mac osx is different from on linux.
- Change coro_swapcontext() definition. The original definition has `symbol not found` issue on mac osx when linking.


lpereira:
Merged with slight changes. Thanks!

LampmanYao:
@lpereira I am sorry for that i introduce a bug in this PR: change the coro_swapcontext() definition will leads to core-dump. But I'm working on this. I will make a new PR when i fixed it.

LampmanYao:
Here's the patch file. Sorry again via this way not a PR. :(
[patch.txt](https://github.com/lpereira/lwan/files/285972/patch.txt)

lpereira:
That's OK. I've applied the patch on HEAD. Thanks again!

PR #146, opened 2016-05-30
Use EV_CLEAR flag when add event to kqueue

LampmanYao:
It fixes the `"UNKNOWN (null) HTTP/1.1" 408 text/html` problem.


lpereira:
Merged, thanks!

PR #148, opened 2016-06-02
Implement the work for a shared library

### LINKED ISSUE #138: Shared Library
It would be great if liblwan-common could be built as a shared library instead of a dynamic library. It would make life a lot simpler for creating arbitrary servers/sites (since the headers and library could distributed and ``#include`` and ``-l`` is all that would be necessary).

HalosGhost:
One of the things I'm most excited about for lwan is to be able to install it as a library system-wide so that I can write third-party code that links to it and uses the web-server as-needed.

This PR will serve to hold the implementations necessary to realize that goal. This should not be merged until @lpereira and I have signed off on its integrity and completeness.

When merged, this will close #138.

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -93,7 +93,8 @@ else ()
 	message(STATUS "Building with Lua support using ${LUA_LIBRARIES}")
 endif ()
 
-add_library(lwan-common STATIC ${SOURCES})
+add_library(lwan-common        SHARED ${SOURCES})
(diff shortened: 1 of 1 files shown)


lpereira:
I've merged this patch, thanks!

I've made a few changes on top of it. It should fix the generation of pkg-config files, and install `liblwan.so` and `liblwan.a`. Please test and tell me if it's working properly, so that #138 can be closed.

HalosGhost:
@lpereira, looks great!

PR #153, opened 2016-06-14
cleaning up the lwan binary

HalosGhost:
This is

1. quite a bit cleaner (less line noise, uses string literal concatenation for simplicity, etc.)
2. should be quite a bit faster since it should avoid multiple, redundant function calls.

### DIFF
--- a/lwan/main.c
+++ b/lwan/main.c
@@ -57,22 +57,22 @@ parse_args(int argc, char *argv[], lwan_config_t *config, char *root)
             result = ARGS_SERVE_FILES;
             break;
 
-        case 'h':
-            printf("Usage: %s [--root /path/to/root/dir] [--listener addr:port]\n", argv[0]);
-            printf("\t[--config]\n");
-            printf("Serve files through HTTP.\n\n");
-            printf("Defaults to listening on %s, serving from ./wwwroot.\n\n", config->listener);
-            printf("Options:\n");
-            printf("\t-r, --root      Path to serve files from (default: ./wwwroot).\n");
-            printf("\t-l, --listener  Listener (default: %s).\n", config->listener);
-            printf("\t-h, --help      This.\n");
-            printf("\n");
-            printf("Examples:\n");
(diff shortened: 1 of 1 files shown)


HalosGhost:
I would still like to go through the rest of ``main.c`` and clean it up a bit more before a merge.

lpereira:
Thanks for the PR! However, I don't like it:

- Calls to `printf()` with large formatting strings are harder to read, as one has to count the positional arguments.
- Since `stdout` is buffered inside stdio, this ends up being about the same thing, performance-wise.
- Breaking up in multiple calls to `printf()` lets one `#ifdef ... #endif` if options end up being enabled or disabled through compile-time flags.
- Compilers will often give out warnings if a variable is passed as the first argument to a `printf()-like` function, for security concerns. These concerns do not apply here as the string isn't user-controllable, but I'm not sure if the compiler will complain or not.

HalosGhost:
* Actually, the next thing I planned to do was remove a few of the positional arguments (namely, there is a line at the top of the usage that says default arguments even though defaults are specified in the option list), and I agree; having to do a lot of that is troublesome.
* It's true there is probably only minimal, if any, performance benefit here.
* Because the change uses string concatenation, you can still do ``#ifdef…#endif`` barring
* That is true, and it could be alleviated by using a macro, or by avoiding the variable and still having the literal in the first argument of ``printf()`` (which is something I would likely do anyway).

Would you accept changes above that would help alleviate some of your concerns, or would you rather not accept this change and I focus on other cleanup efforts?

(as a related side-note: do you care about strict compilation with no warnings? e.g., using ``-Weverything`` will enable quite a few more warnings if ``clang`` is used as the compiler)

lpereira:
I'd prefer if you focused on other cleanup efforts. I'm used to the code so it's pretty hard for me to point you at a certain direction. Maybe focus on implementing the `--config` option?

I like a warning-free compilation, yes, but `-Weverything` from Clang is enables way too many useless ones.

HalosGhost:
Sounds good to me!

PR #164, opened 2016-07-23
Remove missing.h from cmake install targets

escondida:
make install errors out since the removal of missing.h


lpereira:
Merged, thanks!

escondida:
No, thanks you!

PR #165, opened 2016-07-27
HTTP Methods

0x003e:
**Patch HTTP POST request bug** : https://github.com/lpereira/lwan/pull/165/files#diff-c2cfb924f8ec9c2c56b9e8c654f82773L783

 -   Add method PUT, DELETE, OPTIONS and PATCH.

### DIFF
--- a/common/lwan-request.c
+++ b/common/lwan-request.c
@@ -89,9 +89,9 @@ union proxy_protocol_header {
 
 static char decode_hex_digit(char ch) __attribute__((pure));
 static char *ignore_leading_whitespace(char *buffer) __attribute__((pure));
-static lwan_request_flags_t get_http_method(const char *buffer) __attribute__((pure));
+static lwan_request_method_t get_http_method(const char *buffer) __attribute__((pure));
 
-static ALWAYS_INLINE lwan_request_flags_t
+static ALWAYS_INLINE lwan_request_method_t
 get_http_method(const char *buffer)
 {
     /* Note: keep in sync in identify_http_method() */
@@ -99,10 +99,10 @@ get_http_method(const char *buffer)
         HTTP_STR_GET     = MULTICHAR_CONSTANT('G','E','T',' '),
         HTTP_STR_HEAD    = MULTICHAR_CONSTANT('H','E','A','D'),
         HTTP_STR_POST    = MULTICHAR_CONSTANT('P','O','S','T'),
-        HTTP_STR_PUT     = MULTICHAR_CONSTANT_SMALL('P','U'),// PUT
-        HTTP_STR_DELETE  = MULTICHAR_CONSTANT('D','E','L','E'), // DELETE
-        HTTP_STR_OPTIONS = MULTICHAR_CONSTANT('O','P','T','I'), // OPTION
-        HTTP_STR_PATCH   = MULTICHAR_CONSTANT('P','A','T','C') // PATCH
+        HTTP_STR_PUT     = MULTICHAR_CONSTANT('P','U','T',' '),
+        HTTP_STR_DELETE  = MULTICHAR_CONSTANT('D','E','L','E'),
+        HTTP_STR_OPTIONS = MULTICHAR_CONSTANT('O','P','T','I'),
+        HTTP_STR_PATCH   = MULTICHAR_CONSTANT('P','A','T','C')
     };
 
     STRING_SWITCH(buffer) {
@@ -305,8 +305,8 @@ identify_http_method(lwan_request_t *request, char *buffer)
         [REQUEST_METHOD_OPTIONS] = sizeof("OPTIONS ") - 1,
         [REQUEST_METHOD_PATCH] = sizeof("PATCH ") - 1,
     };
-    lwan_request_flags_t flags = get_http_method(buffer);
-    request->flags |= flags;
+    lwan_request_method_t flags = get_http_method(buffer);
+    request->method |= flags;
     return buffer + sizes[flags];
 }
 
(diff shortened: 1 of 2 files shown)


common/lwan-request.c
```diff
@@ -780,18 +796,9 @@ static lwan_read_finalizer_t read_request_finalizer(size_t total_read,
         return FINALIZER_DONE;
     }
 
-    if (LIKELY(!memcmp(helper->buffer->value + total_read - 4, "\r\n\r\n", 4)))
+    if (LIKELY(!memcmp(helper->buffer->value + total_read - 2, "\r\n", 2)))
         return FINALIZER_DONE;
 
-    if (get_http_method(helper->buffer->value) == REQUEST_METHOD_POST) {
-        char *post_data_separator = memrchr(helper->buffer->value, '\n',
```
0x003e: I have a bug when I test with POSTMAN Rest client (Google Chrome),.
Because if the server receive an HTTP POST request ,the to char at the end of body datas is '**\r\n**' not **'\r\n\r\n**'.
The last code code cause a bug: the post request are not handled and the connexion with the client are not close.

common/lwan-request.c
```diff
@@ -780,18 +796,9 @@ static lwan_read_finalizer_t read_request_finalizer(size_t total_read,
         return FINALIZER_DONE;
     }
 
-    if (LIKELY(!memcmp(helper->buffer->value + total_read - 4, "\r\n\r\n", 4)))
```
0x003e: I have a bug when I test with POSTMAN Rest client (Google Chrome),.
Because if the server receive an HTTP POST request ,the to char at the end of body datas is '\r\n' not '\r\n\r\n'.
The last code code cause a bug: the post request are not handled and the connexion with the client are not close.

common/lwan-request.c
```diff
@@ -99,6 +99,10 @@ get_http_method(const char *buffer)
         HTTP_STR_GET     = MULTICHAR_CONSTANT('G','E','T',' '),
         HTTP_STR_HEAD    = MULTICHAR_CONSTANT('H','E','A','D'),
         HTTP_STR_POST    = MULTICHAR_CONSTANT('P','O','S','T'),
+        HTTP_STR_PUT     = MULTICHAR_CONSTANT_SMALL('P','U'),// PUT
```
lpereira: `MULTICHAR_CONSTANT_SMALL` should only be used with `STRING_SWITCH_SMALL`. You can use `MULTICHAR_CONSTANT('P', 'U', 'T', ' ')` here.

Also, there's no need to add comments.

common/lwan.h
```diff
@@ -160,7 +160,11 @@ typedef enum {
     RESPONSE_NO_CONTENT_LENGTH = 1<<8,
     RESPONSE_URL_REWRITTEN     = 1<<9,
     REQUEST_ALLOW_PROXY_REQS   = 1<<10,
-    REQUEST_PROXIED            = 1<<11
+    REQUEST_PROXIED            = 1<<11,
+    REQUEST_METHOD_PUT         = 1<<12,
```
lpereira: Since more methods are being added, it might be a better idea to use `ceil(log2(n_methods))` bits to store the method information. While doing that, it might be a good idea to move the `REQUEST_METHOD_` items to the beginning of this enum, so that the array in `identify_http_method()` is smaller.

For instance, there's already GET, HEAD, and POST methods. By adding PUT, DELETE, OPTIONS, and PATCH, there would be 7 methods. You need only 3 bits to store all of these, since these options are all mutually exclusive. So having something like `REQUEST_METHOD_BITMASK0 = 1<<0, REQUEST_METHOD_BITMASK1 = 1<<1, REQUEST_METHOD_BITMASK2 = 1<<2`, and then building other methods by OR-ing with these bitmasks, you'll save a lot of space in that array: GET can be 001 (`REQUEST_METHOD_BITMASK2`), HEAD can be 010 (`REQUEST_METHOD_BITMASK1`), POST can be 011 (`REQUEST_METHOD_BITMASK1 | REQUEST_METHOD_BITMASK2`), etc.

common/lwan-request.c
```diff
@@ -780,18 +796,9 @@ static lwan_read_finalizer_t read_request_finalizer(size_t total_read,
         return FINALIZER_DONE;
     }
 
-    if (LIKELY(!memcmp(helper->buffer->value + total_read - 4, "\r\n\r\n", 4)))
+    if (LIKELY(!memcmp(helper->buffer->value + total_read - 2, "\r\n", 2)))
         return FINALIZER_DONE;
 
-    if (get_http_method(helper->buffer->value) == REQUEST_METHOD_POST) {
```
lpereira: Are the tests passing after this is removed? Can you provide a test case that requires this to be removed?

common/lwan-request.c
lpereira: Can you please provide the raw request that's being sent by this client? The `requests` Python module seems to work fine with Lwan, and that's what I'm using for tests.

lpereira:
Hi! Thanks for the pull request. I've commented in the patch itself the places I have questions before accepting it.

In addition to these questions, I'd like to ask you a few things:

- Please split this in 3 different patches: 1 adding the new methods, 1 for each part modified in the finalizer function. Splitting the patches makes it easier to use git-bisect to find regressions and/or bugs afterwards, and since there are changes in sensitive parts of the code, this is important.
- It would be interesting if there were test cases for the new methods, just to ensure they're being accepted correctly (some of them are larger than "POST" or "GET", so this might be a problem). Writing a test should be easy (just edit `tools/testsuite.py`), but if you need help, feel free to ask.

common/lwan-request.c
0x003e: x-www-form-urlencoded not have \r\n at the end
aa

form-data have \r\n at the end and not \r\n\r\n
bb

common/lwan.h
0x003e: Ok I try create a special enum for HTTP Method with log2 (number of methods)

common/lwan-request.c
0x003e: https://github.com/lpereira/lwan/pull/165/files#r72793337

lpereira:
Any reason you ended up closing this PR, @0x003e?

0x003e:
Dont worry, I intend to make another "clean" pull request and commits with your recommandations.

0x003e:
I prefer create tests target in CMake before start modification to tract somme regression.

PR #166, opened 2016-07-30
Useful make targets testing and upgrade README

0x003e:


CMakeLists.txt
```diff
@@ -227,3 +227,23 @@ configure_file(
...
+				DEPENDS ${DEPEND}
+				WORKING_DIRECTORY ${WORKING_DIRECTORY}
+				COMMENT "Running Python tests: ${FILE}.")
+	endfunction()
+
+	add_python_test( testsuite
+			${PROJECT_SOURCE_DIR}/tools/testsuite.py
+			${PROJECT_SOURCE_DIR}
+			testrunner)
+
```
0x003e: ~/lwan/build$ make teststuite

CMakeLists.txt
```diff
@@ -227,3 +227,23 @@ configure_file(
...
+	add_python_test( benchmark
+			${PROJECT_SOURCE_DIR}/tools/benchmark.py
+			${PROJECT_SOURCE_DIR}
+			testrunner)
+endif()
```
0x003e: ~/lwan/build$ make benchmark

README.md
```diff
@@ -140,6 +145,18 @@ This will generate a few binaries:
  - `techempower/techempower`: Code for the Techempower Web Framework benchmark. Requires SQLite and MySQL libraries.
  - `common/mimegen`: Builds the extension-MIME type table. Used during build process.
 
+### Tests
+
+    ~/lwan/build$ make teststuite
+
+This will compile `testrunner/testrunner` and execute non regression tests in python file `~/lwan/build/tool/testsuite.py`.
```
lpereira: Minor: "execute regression test suite in `tool/testsuite.py`". The path `~/lwan` considers Lwan is checked out directly in the home directory of a user, which isn't always the case (I clone here in `~/git`).

README.md
```diff
@@ -140,6 +145,18 @@ This will generate a few binaries:
...
+
+### Benchmark
+
+    ~/lwan/build$ make benchmark
+
+This will compile `testrunner/testrunner` and execute benchmark tests in python file `~/lwan/build/tool/benchmark.py`.
```
lpereira: Minor: "execute benchmark script `tools/benchmark.py`."

Same comment about the path applies here.

CMakeLists.txt
```diff
@@ -227,3 +227,23 @@ configure_file(
 )
 
 install(FILES "${CMAKE_BINARY_DIR}/${PROJECT_NAME}.pc" DESTINATION lib/pkgconfig)
+
+if (LUA_FOUND)
+	function(add_python_test NAME FILE WORKING_DIRECTORY DEPEND)
```
lpereira: I'd prefer if the tests failed if Lua were not found rather than only creating the custom target if Lua libraries are present.

tools/benchmark.py
```diff
@@ -171,3 +186,4 @@ def log(self, keep_alive, n_connections, rps, kbps, _2xx, _3xx, _4xx, _5xx):
       sleepwithstatus('Waiting for keepalive connection timeout', keep_alive_timeout * 1.1)
 
   output.footer()
+  lwan.kill()
```
lpereira: Please end the file with a new line.

lpereira:
Thanks for the pull request, @0x003e! Just a few minor nits, however.

CMakeLists.txt
0x003e: > "**This way, other tests that do not depend on Lua will be executed as well.**

Future other tests ? 
Because your actuel "*testrunner*" test depend of Lua because the configuration **testrunner.conf** ligne 54 and 63 have 2 HTTP routes executed with lua.

You want I remove this Lua check ?

lpereira: Indeed, you're right. Nevermind, then.

README.md
0x003e: > which isn't always the case (I clone here in ~/git).

Of course, I have been following your reference path used in Build section of README.

0x003e:
I **amend commit** your minor recommendation.
Now I focus on:

- 1 Create enum **lwan_method_t** with HTTP method GET, POST, PUT, DELETE, PATCH, OPTIONS.
- 2 Create attribute *method** in *lwan_request_t_*.
- 3 Parse GET, POST, PUT, DELETE, PATCH, OPTIONS get_http_method.
- 4 Put the parsed method in atribute **method** in *request* on *identify_http_method* function.

lpereira:
Nice. I'll merge this PR soon. Thanks again!

As for the other PR, I have a few comments I'd like to make before you continue your work. For instance, creating a `lwan_method_t` is pretty nice, but I'd rather have a function that returns a method given a a `lwan_request_t` and do all the bitmasks inside this function, rather than adding another member to the `lwan_request_t` structure.

The flags member is there to save some space and reduce pressure on the cache. Another important thing to keep in mind for cache effects is the `sizes[]` array inside `identify_http_method()`: having a large array there (in other words: a large number of bits to determine the request method) will negatively impact performance.

Also, there's no need to have `enum lwan_method_t` with entries with just 1 bit set, since they're mutually exclusive (it's slightly like this right now inside the flags enum since there are only 3 supported methods).

lpereira:
Merged, thanks!

0x003e:
> I'd rather have a function that returns a method given a a lwan_request_t and do all the bitmasks inside this function, rather than adding another member to the lwan_request_t structure.

#### Do you think the flowing code it's a correct implementation ?
> **My goal is to allow lwan handle HTTP REST request (API).**

log2 (n° lwan_request_flags_t) = 4.
```C
typedef enum {
    REQUEST_ALL_FLAGS          = -1,
    REQUEST_METHOD_GET         = 1<<0,
    REQUEST_METHOD_HEAD        = 1<<1,
    REQUEST_METHOD_POST        = 1<<2,
    REQUEST_METHOD_PUT         = 1<<3,
    REQUEST_METHOD_PATCH       = 1<<4,
    REQUEST_METHOD_DELETE      = 1<<5,
    REQUEST_METHOD_OPTIONS      = 1<<6,
    REQUEST_ACCEPT_DEFLATE     = 1<<7,
    REQUEST_ACCEPT_GZIP        = 1<<8,
    REQUEST_IS_HTTP_1_0        = 1<<9,
    RESPONSE_SENT_HEADERS      = 1<<10,
    RESPONSE_CHUNKED_ENCODING  = 1<<11,
    RESPONSE_NO_CONTENT_LENGTH = 1<<12,
    RESPONSE_URL_REWRITTEN     = 1<<13,
    REQUEST_ALLOW_PROXY_REQS   = 1<<14,
    REQUEST_PROXIED            = 1<<15
} lwan_request_flags_t;
```


```C
static ALWAYS_INLINE lwan_request_flags_t
get_http_method(const char *buffer)
{
    /* Note: keep in sync in identify_http_method() */
    enum {
        HTTP_STR_GET     = MULTICHAR_CONSTANT('G','E','T',' '),
        HTTP_STR_HEAD    = MULTICHAR_CONSTANT('H','E','A','D'),
        HTTP_STR_POST    = MULTICHAR_CONSTANT('P','O','S','T'),
        HTTP_STR_PUT     = MULTICHAR_CONSTANT('P','U','T',' '),
        HTTP_STR_PATCH   = MULTICHAR_CONSTANT('P','A','T','C'),
        HTTP_STR_DELETE  = MULTICHAR_CONSTANT('D','E','L','E'),
        HTTP_STR_OPTIONS = MULTICHAR_CONSTANT('O','P','T','I'),
    };

    STRING_SWITCH(buffer) {
    case HTTP_STR_GET:
        return REQUEST_METHOD_GET;
    case HTTP_STR_HEAD:
        return REQUEST_METHOD_HEAD;
    case HTTP_STR_POST:
        return REQUEST_METHOD_POST;
    case HTTP_STR_PATCH:
        return REQUEST_METHOD_PATCH;
    case HTTP_STR_PUT:
        return REQUEST_METHOD_PUT;
    case HTTP_STR_DELETE:
        return REQUEST_METHOD_DELETE;
    case HTTP_STR_OPTIONS:
        return REQUEST_METHOD_OPTIONS;
    }

    return 0;
}
```

```C
static ALWAYS_INLINE char *
identify_http_method(lwan_request_t *request, char *buffer)
{
    static const char sizes[] = {
        [0] = 0,
        [REQUEST_METHOD_GET]     =  sizeof("GET ")      - 1,
        [REQUEST_METHOD_HEAD]    =  sizeof("HEAD ")     - 1,
        [REQUEST_METHOD_POST]    =  sizeof("POST ")     - 1,
        [REQUEST_METHOD_PUT]     =  sizeof("PUT ")      - 1,
        [REQUEST_METHOD_PATCH]   =  sizeof("PATCH ")    - 1,
        [REQUEST_METHOD_DELETE]  =  sizeof("DELETE ")   - 1,
        [REQUEST_METHOD_OPTIONS] =  sizeof("OPTIOONS ") - 1,
    };
    lwan_request_flags_t flags = get_http_method(buffer);
    request->flags |= flags;
    return buffer + sizes[flags];
}
```

0x003e:
***Execution of the testsuite after***

selection_013

0x003e:
@lpereira

lpereira:
The `log2(n_methods)` must be in the number of bits necessary to store which method is being used. Since they're mutually-exclusive, the best idea is to have a bitmask.

With the proposed approach, the `sizes[]` array in `identify_http_method()` will have `1<<7` (128) elements instead of 3 (`ceil(log2(128))`). Since each will occupy 4 bytes, this table will be about 512 bytes long, using 8 cache lines. If only 3 bits are used for all the 7 methods, the table will have 32 bytes, using only one cache line.

0x003e:
@lpereira 
> the sizes[] array in identify_http_method() will have 1<<7 (128) elements

```C
sizeof("GET ")-1        =       00000100[REQUEST_METHOD_GET]                                                                         
sizeof("HEAD ")-1       =       00000101[REQUEST_METHOD_HEAD]                                                                          
sizeof("POST ")-1       =       00000101[REQUEST_METHOD_POST]                                                                          
sizeof("PUT ")-1        =       00000100[REQUEST_METHOD_PUT]                                                                          
sizeof("PATCH ")-1      =       00000110[REQUEST_METHOD_PATCH]                                                                          
sizeof("OPTIONS ")-1    =       00001000[REQUEST_METHOD_OPTONS]  
```
>  Since each will occupy 4 bytes, this table will be about 512 bytes long, using 8 cache lines. 

I beginner in low level programming, I don't understand why  **this table will be about 512 bytes long**.

>  only 3 bits are used for all the 7 methods

How do you implement that in C ?
Can you give me a example or make a commit with this implementation ?

Thank you for sharing of your skills and time.

lpereira:
Alright. Let me exemplify.

The syntax `1<<n` means "take the integer 1 and shift it left by n bits". It ends up being equivalent to `2^n`. So `1<<7` is 128. When using [designated initializers](https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html) in C, it's possible to initialize only a few elements. So, for instance:

    int array[] = {
         [1<<0] = 1,
         [1<<2] = 2,
         [1<<10] = 3
    };

This will generate an array of `1<<10 + 1` (1025) elements. All of them, with the exception of elements 1, 4, and 1024, will be set to `0`. These other elements will be set to `1`, `2`, and `3`, respectively.

Since each integer is 4 bytes long, an array with `1<<7 + 1` elements will have `(128 + 1) * 4` bytes, so actually 516 bytes. Since each cache line is usually 64-bytes, that's 9 cache lines.

So, the idea is to create a bitmask in the flags enum:

    enum {
       FLAGS_METHOD_GET = 1<<0,
       FLAGS_METHOD_POST = 1<<1,
       FLAGS_METHOD_HEAD = 1<<0 | 1<<1,
       FLAGS_METHOD_DELETE = 1<<2,
       FLAGS_METHOD_OPTIONS = 1<<2 | 1<<0,
       FLAGS_METHOD_PUT = 1<<2 | 1<<1,
       ...
       FLAGS_METHOD_MASK = 1<<0 | 1<<1 | 1<<2
    };

This way, the array will be at most `4 * (1<<2 + 1)`, or 20 bytes long. Then, to determine which method given a set of flags:

    lwan_request_method_t lwan_request_get_method(const lwan_request_t *req)
    {
          static const enum methods[] = {
              [FLAGS_METHOD_GET] = REQUEST_METHOD_GET,
              [FLAGS_METHOD_POST] = REQUEST_METHOD_POST,
              [FLAGS_METHOD_HEAD] = REQUEST_METHOD_HEAD,
              ...
          };
          return methods[req->flags & FLAGS_METHOD_MASK];
    }

The operation `& FLAGS_METHOD_MASK` only selects the bits used to store which method a request is using.

If it's still not clear, feel free to ask again.

0x003e:
OK OK OK !

It's clear, I come from C++ I not familiar with the C99 syntax.

If the test of this implementation (optimization) are correct, I think that It's a good subject for your blog :+1: .

If I have time, I will make a benchmark to estimate the gains of cache performance with a another tool than weighttp (because it's support only GET method).

Now I follow your recommendations for a future pull request.

0x003e:
(Beside the point):

This last week-end, I made a benchmark of LWAN vs GWAN(http://gwan.com/), and I very suprise of the result because it's  almost the same !



```
$ cat /proc/meminfo | grep MemTotal                                                
MemTotal:        8084180 kB

$ lscpu | grep "Model name"
Model name:            Intel(R) Core(TM) i5-2500K CPU @ 3.30GHz
```

**http://localhost:8080/100.html**
>**Benchmark configuration**: Step 250, Start at 10 client, End at 100K client,
Thread 2, Request 1 000K. 

# GWAN
gwan

# LWAN
lwan-step250-startcon10-endcon100k-tread2-request1m

0x003e:
Hi @lpereira,
I try to implement PUT, PATCH, DELETE, OPTIONS method and I failed.

> the idea is to create a bitmask in the flags enum:

**so in this enum :**
```C
typedef enum {
    REQUEST_ALL_FLAGS          = -1,
    REQUEST_METHOD_GET         = 1<<0,
    REQUEST_METHOD_HEAD        = 1<<1,
    REQUEST_METHOD_POST        = 1<<2,
    REQUEST_ACCEPT_DEFLATE     = 1<<3,
    REQUEST_ACCEPT_GZIP        = 1<<4,
    REQUEST_IS_HTTP_1_0        = 1<<5,
    RESPONSE_SENT_HEADERS      = 1<<6,
    RESPONSE_CHUNKED_ENCODING  = 1<<7,
    RESPONSE_NO_CONTENT_LENGTH = 1<<8,
    RESPONSE_URL_REWRITTEN     = 1<<9,
    REQUEST_ALLOW_PROXY_REQS   = 1<<10,
    REQUEST_PROXIED            = 1<<11
} lwan_request_flags_t;
```
I have a difficulty with the `enum lwan_request_flags_t`, because if:
> the array will be at most 4 * (1<<2 + 1), or 20 bytes long. Then, to determine which method given a set of flags:

What is the value of the **FLAGS_METHOD_MASK** for 4 more methods we need to implement  PUT, PATCH, DELETE, OPTIONS (eventual access control origin) to make a REST API ?

*Because REQUEST_METHOD_OPTIONS have the value: 1<<2 | 1<<1 | 1<<0,* ***FLAGS_METHOD_MASK should be 1<<3 | 1<<2 | 1<<1 | 1<<0*** *?*

selection_015

***

The second difficulty depend of the first: 
when you use **if** to determine the METHOD_REQUEST_**?** of the **request->flag**, I think it's necessary have a the mask to compare in **if** statement:
selection_014

***

If you could answer me that directly (when you not busy)  **with a commit of your solution** (or the source code), I would be most grateful; and I could be write the python tests of the all method handled by lwan.

PR #174, opened 2016-10-12
Remove warning on 32 bit ARM build

hdante:
This patch works around a gcc deficiency that evaluates all possible
architecture specific code paths, instead of only evaluating the correct
one, resulting in warnings on the incorrect code paths.

Results in identical code, confirmed with sha512sum, in the following build environments:

- gcc + arm
- clang + arm
- gcc + amd64
- clang + amd64


lpereira:
Merged, thanks!

PR #178, opened 2016-10-31
Support for cross-domain requests

fischermario:
A word of caution: this is officially my first® pull request on GitHub. If I did something wrong or if I messed this up completely, please stay nice ;)

What is this? Sometimes there is the need to request data from a server on a different domain or a different machine on a local network. This may be done by using Javascript (e.g. jQuery) and is considered a 'cross-domain request', which is usually not allowed. To allow such kind of requests the webserver has to explicitly do it by sending additional information inside the response header. What makes things worse is the browser doing a 'preflight request' by performing an additional request using the request method 'OPTIONS' before issuing the actual request to the desired resource.

As this is a very convoluted topic, I did some research about CORS (Cross-Origin Resource Sharing). There is good documentation [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html) (sec. 9.2) and [here](http://enable-cors.org/server.html).

The two patches do not implement the full specification. They are merely a starting point. Just put "allow_cors = yes" in the top level of the configuration file (outside the listener section) to enable CORS support. That is enough to make things work on my end.

Of course there could be many more options and additional logic, but remember that this is a rather special use case and I would like some feedback on the topic first.

### DIFF
--- a/common/lwan-response.c
+++ b/common/lwan-response.c
@@ -105,6 +105,8 @@ get_request_method(lwan_request_t *request)
         return "HEAD";
     if (request->flags & REQUEST_METHOD_POST)
         return "POST";
+    if (request->flags & REQUEST_METHOD_OPTIONS)
+        return "OPTIONS";
     return "UNKNOWN";
 }
 
@@ -173,7 +175,7 @@ lwan_response(lwan_request_t *request, lwan_http_status_t status)
         return;
     }
 
-    if (request->flags & REQUEST_METHOD_HEAD) {
+    if (request->flags & (REQUEST_METHOD_HEAD | REQUEST_METHOD_OPTIONS)) {
         lwan_write(request, headers, header_len);
         return;
     }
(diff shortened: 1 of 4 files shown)


common/lwan.c
```diff
@@ -401,6 +402,9 @@ static bool setup_from_config(lwan_t *lwan, const char *path)
             } else if (streq(line.line.key, "proxy_protocol")) {
                 lwan->config.proxy_protocol = parse_bool(line.line.value,
                             default_config.proxy_protocol);
+            } else if (!strcmp(line.line.key, "allow_cors")) {
```
lpereira: I'd use `streq()` here instead.

common/lwan-response.c
```diff
@@ -313,6 +317,13 @@ lwan_prepare_response_header(lwan_request_t *request, lwan_http_status_t status,
         APPEND_STRING_LEN(request->conn->thread->date.expires, 29);
     }
 
+    if (request->conn->thread->lwan->config.allow_cors) {
```
lpereira: Not really necessary, but maybe minimize this indirection by introducing a `REQUEST_ALLOW_CORS` flag and setting it inside `process_request_coro()`, since there's a pointer to a `lwan_t` struct in that function already?

lpereira:
I went ahead and merged this PR with the changes that have been requested. Thanks, CORS request headers ended up being useful for one of my use cases!

PR #185, opened 2017-02-27
lwan return type fix for C++/C compatability

GuacheSuede:


lpereira:
Thanks for the contribution, but I'm closing this issue as it's a duplicate of #186.

GuacheSuede:
#186 does not work when compiled, does it work for you ?

lpereira:
Changing the return type is not the ideal choice: an int doesn't have an attached meaning, and if an enum can be used instead, it's better. In fact, a better fix would be converting from the request flags to another enum that contains only the request methods (although that requires changing things in yet another place when a new request method is added).

And, no, #186 doesn't work with a C compiler, but I've asked the author to use a different kind of cast that will work with both C++ and C.

GuacheSuede:
C type casting does not work in C++ in that instance.

Regarding your suggesting fix, what kind of implementation other than enum do you suggest, member functions ?

lpereira:
Lwan is written in C, so it doesn't have member functions. If C type casting won't work in C++ in that case, maybe the best way to do it is to move the implementation to lwan-request.c and have only the prototype in the .h file.

GuacheSuede:
Thanks @lpereira , Compilation works now.

PR #186, opened 2017-02-27
Fix compilation with C++

matt-42:


common/lwan.h
```diff
@@ -382,7 +382,7 @@ void lwan_format_rfc_time(time_t t, char buffer[ENFORCE_STATIC_BUFFER_LENGTH 30]
 static inline enum lwan_request_flags
 lwan_request_get_method(const struct lwan_request *request)
 {
-    return request->flags & REQUEST_METHOD_MASK;
+    return lwan_request_flags(request->flags & REQUEST_METHOD_MASK);
```
lpereira: Please use a C-style cast, as this won't work with C. (I don't use a C++ compiler for Lwan.)

lpereira:
Looks good -- but please squash all changes into one and push again. This way the history is cleaner.

PR #192, opened 2017-03-26
Add another happy and grateful user of Lwan :)

friedcalamari:
Added KrakenD in the Readme


lpereira:
Merged as a single commit. Thanks!

PR #195, opened 2017-04-27
Fix coredump on OS X

LampmanYao:


CMakeLists.txt
```diff
@@ -132,8 +136,13 @@ else ()
 		unset(CMAKE_REQUIRED_FLAGS)
 
 		if(HAVE_FLAG_SANITIZE_UNDEFINED)
-			message(STATUS "Build with newer undefined behavior sanitizer")
-			set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fsanitize=undefined")
+			if (APPLE)
```
lpereira: I'm not sure if I understand this change in particular.  Undefined Behavior sanitizer can be disabled by passing `-DUBSAN=OFF` to the CMake configuration line, and it should work for any supported platform. Address Sanitizer can be enabled by passing `-DASAN=ON`.  This should ideally be changed to `-DSANITIZE={none,ub,address}` or something like that, though.

CMakeLists.txt
```diff
@@ -105,7 +105,11 @@ if (${CMAKE_BUILD_TYPE} MATCHES "Rel")
 
 	check_c_compiler_flag(-flto HAS_LTO)
 	if (HAS_LTO)
-		set(C_FLAGS_REL "${C_FLAGS_REL} -flto")
+		if (APPLE)
```
lpereira: I'm not sure if LTO is indeed an issue.  The crash reported by @jwoertink also appears to happen in debug builds, which are not built with LTO enabled.

LampmanYao:
I closed this PR, since it can be compiled with `-DASAN=ON`.

PR #210, opened 2017-10-19
Do not print colors on dumb terminals

escondida:
Trivial patch: test for the catch-all "dumb" terminal before printing colors. It's conventional for oddball terminals that don't do ANSI colors to set TERM to "dumb".

### DIFF
--- a/src/lib/lwan-status.c
+++ b/src/lib/lwan-status.c
@@ -43,6 +43,8 @@ static volatile bool quiet = false;
 static bool use_colors;
 static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
 
+static bool can_use_colors(void);
+
 void
 lwan_status_init(struct lwan *l)
 {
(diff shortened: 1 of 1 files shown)


src/lib/lwan-status.c
```diff
@@ -52,7 +52,7 @@ lwan_status_init(struct lwan *l)
     quiet = false;
     (void) l;
 #endif
-    use_colors = isatty(fileno(stdout));
+    use_colors = isatty(fileno(stdout)) && strcmp(getenv("TERM"), "dumb");
```
lpereira: `getenv()` will return NULL if an environment variable isn't set.  Might be best to create a static function, break this into multiple lines, and call it from within `lwan_status_init()`. Also, please use `secure_getenv()` instead:

```C
static bool can_use_colors(void)
{
    const char *term;

    if (!isatty(fileno(stdout))
          return false;

    term = secure_getenv("TERM");
    if (term && streq(term, "dumb"))
            return false;

    return true;
}
```

escondida:
Thanks for the feedback, lpereira!

I've updated the patch, with one difference. Rather than having the static bool function can_use_colors() set the global static bool variable use_colors, I just had the functions that checked use_colors call can_use_colors(). If I guess correctly, the compiler will translate that into basically the same thing, and I figure that the fewer representations of the same data the better.

lpereira:
If you do like this, it is going to call isatty() (which is a syscall) and
getenv() (which does a linear search in the environment variables) for
every status messages that are printed.  Calling it just once in the init()
function will avoid this overhead as it will be called just once.

I can fix this before merging the patch or wait for you to send a revision.

escondida:
Thanks again for getting back to me! Should all be in order now.

lpereira:
Merged, thanks!

PR #212, opened 2017-12-04
Fixes a small memory leak in lwan->config.listener

owah:
The program almost cleans up all of its resources, but in the parse_listener function strdup() is used to overwrite a pointer to an existing piece of malloc'ed memory without freeing it first.


lpereira:
Thanks for finding this -- I'll make a minor adjustment (there's no need to check for NULL pointer before calling free() -- free() does that internally) and push the fix.

owah:
Great! (I actually read that in the documentation a few minutes ago, but I didn't want to bloat my request with additional edits.)

lpereira:
Merged, thanks!

PR #221, opened 2018-02-13
double free config_file_path fix

houseofkodai:


lpereira:
Thanks for the pull request!

The proper fix would be removing the `free(config_file_path)` from src/lib: this way, whoever allocated it has the responsability of freeing it.

Also, please send a single commit in this pull request; there's no need for a merge commit.

PR #222, opened 2018-02-14
remove free(c.listener/config_file_path) in out:

houseofkodai:
since listener/config_file_path memory-allocation occurs after freeing the existing values - the lib/lwan_shutdown is the correct place to free them.

### DIFF
--- a/src/bin/lwan/main.c
+++ b/src/bin/lwan/main.c
@@ -259,8 +259,6 @@ main(int argc, char *argv[])
     lwan_shutdown(&l);
 
 out:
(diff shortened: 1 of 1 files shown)


lpereira:
Thanks for the PR! I ended up solving the problem in another way, by copying the configuration file path in the lwan_config structure (212fe504). This way there's no double free or guessing who owns the pointer.

As far as the listener goes: it's copied in lwan_init_with_config(), and freed in lwan_shutdown(), so it's fine; double freeing that value isn't going to happen.

PR #229, opened 2018-04-04
Cross Compilation Fixes

juntalis:
Encountered a few build issues while using lwan as a test case for a cygwin-hosted x86_64-linux-gnu cross-compiler I'd just built.

* Without the docstring, CMake (at least 3.6.2) sees the ``set(IMPORT_EXECUTABLES "IMPORTFILE-NOTFILE" CACHE FILEPATH)`` as invalid.
* CMake >= 2.6 is capable of resolving mimegen and bin2hex to their corresponding targets. Changed it from the explicit path as a cross compile is likely to happen in a different binary dir.
* The above behavior doesn't seem to apply to imported targets unless explicitly including the `IMPORT_EXECUTABLES` file from the file referencing them.
* Minor tweaks to allow a Cygwin native build for the two generator executables.

There was one other native compilation issue related to the usage of: ``-Wl,-z,now -Wl,-z,relro`` when linking the generator executables. (which Cygwin's ld doesn't recognize and won't ignore) Seemed like a pain to try to resolve, so I just hand-edited the generated link.txt files accordingly. (the [official solution](http://www.cygwin.com/ml/cygwin/2010-01/msg00719.html))

### DIFF
--- a/src/lib/hash.c
+++ b/src/lib/hash.c
@@ -26,7 +26,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
+#ifndef __CYGWIN__
 #include <sys/syscall.h>
+#endif
 #include <sys/types.h>
 #include <unistd.h>
 
--- a/src/lib/hash.c
+++ b/src/lib/hash.c
@@ -26,9 +26,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
-#ifndef __CYGWIN__
 #include <sys/syscall.h>
-#endif
 #include <sys/types.h>
 #include <unistd.h>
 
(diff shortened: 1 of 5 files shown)


src/lib/hash.c
```diff
@@ -26,7 +26,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
+#ifndef __CYGWIN__
```
lpereira: Please create a `syscall.h` file in `src/lib/missing/sys` instead, with the following code:

```C
#ifndef __CYGWIN__
#include_next <sys/syscall.h>
#endif
```

(With the license boilerplate as well.)

This ensures that we don't need to sprinkle #ifdefs everywhere.

juntalis:
Nah, that's reasonable - I should've paid more attention. This last commit should do it.

lpereira:
Merged, thanks!

PR #240, opened 2018-08-25
expose some needed symbols for template authoring

HalosGhost:
It looks like, at least, these 7 symbols need to be exposed in order to author some basic templates (when using ``lwan`` as a dynamic library).

There may be more needed, but I know I need at least these for now.


lpereira:
Indeed these will be needed if you use the templating engine. Merged, thanks

HalosGhost:
Thank you!

PR #241, opened 2018-08-27
Expose an additional symbol for SERVE_FILES() to function

HalosGhost:


lpereira:
Looks like you forgot to rebase the branch on top of head. I've merged your patch with a slight modification (exposing `lwan_module_info_*` instead of only the one for the file serving module, so all modules can be instantiated, even if defined by the user.)

Thanks for testing this -- I don't usually use the shared library.

HalosGhost:
Cheers, thanks! (and yeah, sorry about the rebase mixup)

As I keep adding features to my own things, I imagine I will keep discovering more :)

PR #245, opened 2018-08-29
include stddef and stdbool for this header

HalosGhost:
This is a minor thing; but, if this header is ever used without ``stddef.h`` or ``stdbool.h`` being included, it will throw compile errors.

### DIFF
--- a/src/lib/queue.h
+++ b/src/lib/queue.h
@@ -8,6 +8,9 @@
 
 #pragma once
 
+#include <stddef.h>
+#include <stdbool.h>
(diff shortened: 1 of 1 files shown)


lpereira:
Merged, thanks.

Just be careful when using this queue; this is threadsafe but only when there's just one producer and one consumer.

HalosGhost:
I am personally not using this. I just ran across it as part of my own little autoconf replacement caught it as not having the right includes. Thank you for the warning though!

PR #247, opened 2018-09-03
break LWAN_HANDLER() into declaration and definition

HalosGhost:
Leaving ``LWAN_HANDLER()`` in-place means this should have no affect on existing code.

The benefit is that library users can split declaration from definition (e.g., allow declaring various handlers in a ``.h`` file, but implementing them in a ``.c``). In particular, this enables library users to expose handlers for use in other files.


lpereira:
Merged, thanks!

PR #248, opened 2018-09-08
Fix some errors

okalachev:
Trying to run on macOS High Sierra. Building failed, so I fixed some obvious errors/typos.

With this fixes building succeed. But the server doesn't run printing `Killed: 9` though :)

### DIFF
--- a/src/lib/missing-pthread.c
+++ b/src/lib/missing-pthread.c
@@ -18,6 +18,7 @@
  */
 
 #define _GNU_SOURCE
+#include <errno.h>
 #include <string.h>
 #include <pthread.h>
 
@@ -91,7 +92,7 @@ int pthread_set_name_np(pthread_t thread, const char *name)
 #elif defined(__APPLE__)
 int pthread_set_name_np(pthread_t thread, const char *name)
 {
-    if (!pthread_equal(thread, pthread_self())
+    if (!pthread_equal(thread, pthread_self()))
         return EPERM;
 
     /* macOS takes a char*; I don't know if it's modified or not, so


lpereira:
Merged, thanks!  Now, I can't reproduce the crash.  It works fine on High Sierra here (running 10.13, build 17A365), built with Clang 9.1.0 (clang-902.0.39.2). It just errors out with that signal, or some message is printed out?

okalachev:
It just prints `Killed: 9`:

```
$ src/bin/lwan/lwan
Killed: 9
```

Clang version:

```
$ clang --version
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
```

macOS High Sierra 10.13.6 (17G65).

lpereira:
That's indeed odd. Is this a debug or a release build?

okalachev:
Release.

okalachev:
I'm not very experienced in debugging C programs, but that's what I got after some attempts.

Debug build falls with the same error. Running in GDB says:

```
[New Thread 0x1207 of process 41051]
During startup program terminated with signal SIGKILL, Killed.
```

Fast googling says that this may be related to some linking problems. Also, CMake showed this warning:

```
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   lwan-shared
```

PR #253, opened 2018-10-18
attempt to fix config file warning when using an in-code config

### LINKED ISSUE #249: Allow initialization without config file
At the moment, there are two init functions:

* ``lwan_init()``
* ``lwan_init_with_config()``

As a side-note, the naming convention of these functions is somewhat confusing. The first function implicitly reads from a lwan configuration file (with the same name as the binary being run). The second allows you to pass a configuration struct, and therefore should allow for not needing a configuration file to be present.

Unfortunately, using ``lwan_init_with_config()`` does not escape the need for a configuration file. Lwan still requires a minimal config file when using this function, roughly:

```
listener *:8080 {}
```

Ideally, using ``lwan_init_with_config()`` should escape any need for a configuration file. Perhaps a configuration file, if one is present should still be honored (though I am unsure if that's actually preferable), but it certainly should not be required.

It looks like the specific culprit here is [``try_setup_with_config()``](https://github.com/lpereira/lwan/blob/master/src/lib/lwan.c#L444-L455) (reproduced below for ease):

```c
static void try_setup_from_config(struct lwan *l, const struct lwan_config *config)
{
    if (!setup_from_config(l, config->config_file_path)) {
        if (config->config_file_path) {
            lwan_status_critical("Could not read config file: %s",
                                 config->config_file_path);
        }
    }

    /* `quiet` key might have changed value. */
    lwan_status_init(l);
}
```

I am hoping this can be modified such that if ``config != &default_config``, it skips the check for the config file path. Does that sound reasonable? Are there any assumptions I am making which mean this needs to be done in a different way or that this cannot be done at all?

HalosGhost:
Hopefully fixes #249. At least locally, this works for me; the server runs and no warning is printed.

I would, however, appreciate a second set of eyes on this and some more testing.

### DIFF
--- a/src/lib/lwan-response.c
+++ b/src/lib/lwan-response.c
@@ -25,6 +25,7 @@
 #include <sys/socket.h>
 #include <sys/uio.h>
 #include <unistd.h>
+#include <endian.h>
 
 #include "lwan-private.h"
 
--- a/src/lib/missing/endian.h
+++ b/src/lib/missing/endian.h
@@ -1,0 +1,123 @@
+// This file originates to a work of Mathias Panzenböck, portable_endian.h, of
+// which he puts in public domain.
+//
+// License text was as next.
+//
+// "License": Public Domain
+// I, Mathias Panzenböck, place this file hereby into the public domain.  Use
+// it at your own risk for whatever you like.  In case there are jurisdictions
+// that don't support putting things in the public domain you can also consider
+// it to be "dual licensed" under the BSD, MIT and Apache licenses, if you want
+// to.  This code is trivial anyway.  Consider it an example on how to get the
+// endian conversion functions on different platforms.
+
+#pragma once
+
+#if (defined(_WIN16) || defined(_WIN32) || defined(_WIN64)) &&                 \
+    !defined(__WINDOWS__)
+
+#define __WINDOWS__
+
+#endif
+
+#if defined(__linux__) || defined(__CYGWIN__)
+
+#include_next <endian.h>
+
+#elif defined(__APPLE__)
+
+#include <libkern/OSByteOrder.h>
+
+#define htobe16(x) OSSwapHostToBigInt16(x)
+#define htole16(x) OSSwapHostToLittleInt16(x)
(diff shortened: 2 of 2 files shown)


lpereira:
Thanks for the PR -- unfortunately, I don't think this is the right way to go. The "testrunner" binary, for instance, will be unable to find its configuration file.

HalosGhost:
The updated condition is still non-functional. Will keep at it and let you know when it may be ready for review again.

HalosGhost:
Okay. I have two possibilities that could solve this issue without breaking compatibility:

- Create a separate function that mirrors ``lwan_init_with_config()`` almost exactly save loading a config file
   - (possibly) extract the parts of these two functions to a helper function to avoid repetition and fragility
- Add a boolean flag to the config struct that disables loading a config file
   - modify ``lwan_init_with_config()`` to respect this flag

If the boolean flag route makes more sense, then we would need to decide if it should disable all config file loading or just disable searching for a default config. I.e., it could just disable searching for ``progname.conf`` if ``config->config_file_path`` is ``NULL`` (which is the specific bit of functionality that I am trying to avoid). Obviously then, ``default_config`` would set this value to be ``true`` to preserve backwards compatibility.

Do either of these ideas sound reasonable? Both should hopefully manage to accomplish what I'm looking for without modifying behavior that anyone else is relying on.

lpereira:
So the issue you're facing here is that you're bothered by the warning message that the configuration file can't be opened.  Lwan works fine otherwise.

A simpler solution that do not require changing the API or introducing another initialization function is just changing the warning message to a debugging message, so that it's only displayed on debug builds.  Would that be OK?

HalosGhost:
The warning is part of it, but I'd also like to be able to know that it's not trying to load a config file at all (call me a control-freak if you'd like, :P). But, I should say this is not make-or-break for me. Having the warning not show up does sound pleasant, but even leaving it be would be fine if this seems like a total non-issue :)

HalosGhost:
I am considering this closed pending a better solution (though I'm not worried about it).

PR #258, opened 2018-12-16
add portable_endian.h

7k8m:
Not sure best solution, but putting this header file resolve the compilation error on Mac

### DIFF
--- a/src/lib/missing/endian.h
+++ b/src/lib/missing/endian.h
@@ -8,8 +8,7 @@
 // be "dual licensed" under the BSD, MIT and Apache licenses, if you want to. This code is trivial anyway. Consider it
 // an example on how to get the endian conversion functions on different platforms.
 
-#ifndef PORTABLE_ENDIAN_H__
-#define PORTABLE_ENDIAN_H__
+#pragma once
 
 #if (defined(_WIN16) || defined(_WIN32) || defined(_WIN64)) && !defined(__WINDOWS__)
 
@@ -119,4 +118,3 @@
 
 #endif
 
-#endif


lpereira:
Thanks for the PR, @7k8m!  I rarely test things on a Mac, so PRs like these are welcome.  (I also apologize for not replying sooner; I was on vacations.)

This is almost good for inclusion -- a simple way to get it ready would be renaming `portable_endian.h` to `endian.h`, and moving it to `src/lib/missing`.  Inside, instead of `#include <endian.h>`, just change it to `#include_next <endian.h>`.  Then, in the places where `be64toh()` are used, just `#include <endian.h>` near the header.  How Lwan uses `#include_next` and the `missing` directory (for portability layers such as this) is documented in the `README.md` file, in the "Hacking" section.

7k8m:
Thank you for your response.
I modified to use #include_next. Please confirm.

7k8m:
Ah sorry, I found some mistake ... I will fix it.

7k8m:
I fixed it.
Again, please confirm.

src/lib/missing/endian.h
```diff
@@ -0,0 +1,112 @@
+#ifndef PORTABLE_ENDIAN_H__
```
lpereira: The only thing that's missing here is the licensing information from the original author.  Even though it was placed in the public domain, giving due credit is a nice thing to do.

(While you're at it, use a `#pragma once` instead of the include guard, as mentioned in the [Coding Style](https://github.com/lpereira/lwan#coding-style) section in the README file.)

7k8m:
I applied changes for license texts and use of #pragma once.
Please confirm.

PR #260, opened 2019-01-04
Fix errant Libs line for pkg-config file

### LINKED ISSUE #259: lwan.pc compiled on aarch64 has an errant Libs line
When compiling `lwan` on aarch64, the created `lwan.pc` file adds, along with `-lm`, an `m` entry to the `Libs` line:

```
Name: lwan
Description: Scalable, high performance, experimental web server
Version: 7e212b1
Requires: 
prefix=/usr/local
includedir=${prefix}/include/lwan
libdir=${prefix}/lib
Libs: -L${libdir} -llwan /usr/lib/libz.so -lpthread -llua5.1 m -llua5.1 -lm
Cflags: -I${includedir} -include ${prefix}/include/lwan/lwan-build-config.h
```

Inevitably, as `m` is not a recognized a flag or file, it results in the following error when attempting to link against `lwan` as a dynamic library:

```console
$ gcc `pkg-config --libs lwan` main.c
gcc: error: m: No such file or directory
```

Briefly looking at `grep -rniI 'ADDITIONAL_LIBRARIES'`, it looks like there are only a few places that could be involved with this issue. In particular, it looks like line 45 from the main CMakeLists.txt is the most likely culprit:

```
list(APPEND ADDITIONAL_LIBRARIES "-l${LUA_LIBRARIES} ${LUA_LDFLAGS}")
```

Any thought on how to fix this?

HalosGhost:
Fixes #259


lpereira:
Merged, thanks!

PR #265, opened 2019-04-20
Add necessary header for OpenBSD

sunqingyao:
Otherwise, you will get "error: use of undeclared identifier 'SHUT_RDWR'" on OpenBSD 6.4


lpereira:
Thanks for the PR! Merged.

PR #274, opened 2019-12-05
adding link for nanos support

eyberg:


lpereira:
Thanks for making sure Lwan works on Nanos! Do you have any performance comparison vs running it on Linux?

eyberg:
not yet - we still have some work to do there but i'll def. add it when it exists

PR #276, opened 2020-01-14
Rewrite example: optional forward slash between rewrite base and pattern

curiousleo:
I found it a little surprising that the comment in the example mentions (only) `/some/base/endpointimgur/mp4/4kOZNYX` (no forward slash between "endpoint" and "imgur") whereas the text below talks (only) about `/some/base/endpoint/imgur/mp4/4kOZNYX` (with forward slash between "endpoint" and "imgur"). Both do in fact work, so I think it would be clearer if the example simply listed both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,11 +415,11 @@ example, where two patterns are specified:
 ```
 rewrite /some/base/endpoint {
     pattern posts/(%d+) {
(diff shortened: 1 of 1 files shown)


lpereira:
Oh,  yes, you're right. It might be a good idea to mention anchors there.

PR #280, opened 2020-06-07
(#261) Implement low-cost global-headers

### LINKED ISSUE #261: Add a `lwan_config` field for a set of headers to always serve
Ideally, this would be a `struct lwan_key_value *` member in `struct lwan_config`. Then, every response will unconditionally serve any headers in this member.

This might be really useful for security-related headers that you wish to apply site-wide (for example, `HTTP-Strict-Transport-Security` or `Content-Security-Policy`).

Personally, I would plan to have the following:

```c
static struct lwan_key_value headers [] = {
    { .key = "content-security-policy", .value = "default-src 'self'" },
    { .key = "x-frame-options", .value = "SAMEORIGIN" },
    { .key = "x-xss-protection", .value = "1; mode=block" },
    { .key = "x-content-type-options", .value = "nosniff" },
    { .key = "referrer-policy", .value = "no-referrer" },
    { .key = "strict-transport-security", .value = "max-age=31536000; includeSubDomains" },
    { .key = NULL, .value = NULL }
};
```

HalosGhost:
Includes:

- Defaulting global headers to "Server: lwan"
- Allowing specification of a key-value set for headers in config struct
- No added memory leaks (according to ``valgrind``)
- Basic configuration-file support for the ``headers`` section
    - Specifying a header section in a config file ignores global_headers member of the config struct
    - Does not allow disabling the Server header
- Closes #261

### DIFF
--- a/src/lib/lwan.c
+++ b/src/lib/lwan.c
@@ -132,6 +132,91 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
     return copy;
 }
 
+static void lwan_headers_init(struct lwan *l, struct lwan_key_value *hdrs)
+{
+    assert(l);
+
+    struct lwan_strbuf * headers = &l->headers;
+    lwan_strbuf_init(headers);
+    bool set_server = false;
+    if (hdrs) {
+        for (struct lwan_key_value *kv = hdrs; kv->key; ++kv) {
+            if (!strcasecmp(kv->key, "server"))
+                set_server = true;
+
+            if (LIKELY(kv->value)) {
+                lwan_strbuf_append_printf(headers, "\r\n%s: %s", kv->key, kv->value);
+            }
+        }
+    }
+
+    if (LIKELY(!set_server)) {
+        lwan_strbuf_append_strz(headers, "\r\nServer: lwan");
+    }
+
+    lwan_strbuf_append_strz(headers, "\r\n\r\n");
+}
+
+static void parse_global_headers(struct config *c,
+                                 const struct config_line *l,
+                                 struct lwan *lwan)
+{
+    struct lwan_key_value *kv;
+    struct lwan_key_value_array hdrs;
+    lwan_key_value_array_init(&hdrs);
+
+    while ((l = config_read_line(c))) {
+        switch (l->type) {
+        case CONFIG_LINE_TYPE_SECTION:
+            config_error(
+                c, "No sections are supported under the 'headers' section");
+            goto cleanup;
+
+        case CONFIG_LINE_TYPE_LINE:
+            kv = lwan_key_value_array_append(&hdrs);
+            if (!kv) {
+                lwan_status_critical_perror(
+                    "Could not allocate memory for custom response header");
+            }
+
+            kv->key = strdup(l->key);
+            if (!kv->key) {
+                lwan_status_critical_perror(
+                    "Could not allocate memory for custom response header");
+            }
+
+            kv->value = strdup(l->value);
+            if (!kv->value) {
+                lwan_status_critical_perror(
+                    "Could not allocate memory for custom response header");
+            }
+            break;
+
+        case CONFIG_LINE_TYPE_SECTION_END:
+            kv = lwan_key_value_array_append(&hdrs);
+            if (!kv) {
+                lwan_status_critical_perror(
+                    "Could not allocate memory for custom response header");
+            }
+
+            kv->key = NULL;
+            kv->value = NULL;
+
+            lwan_headers_init(lwan, lwan_key_value_array_get_array(&hdrs));
+            goto cleanup;
+        }
+    }
+
+    config_error(c, "EOF while looking for end of 'headers' section");
+
+cleanup:
+    LWAN_ARRAY_FOREACH (&hdrs, kv) {
+        free(kv->key);
+        free(kv->value);
+    }
+    lwan_key_value_array_reset(&hdrs);
+}
+
 static void parse_listener_prefix_authorization(struct config *c,
                                                 const struct config_line *l,
                                                 struct lwan_url_map *url_map)
@@ -416,6 +501,8 @@ static bool setup_from_config(struct lwan *lwan, const char *path)
                 }
             } else if (streq(line->key, "straitjacket")) {
                 lwan_straitjacket_enforce_from_config(conf);
+            } else if (streq(line->key, "headers")) {
+                parse_global_headers(conf, line, lwan);
             } else {
                 config_error(conf, "Unknown section type: %s", line->key);
             }
@@ -543,6 +630,10 @@ void lwan_init_with_config(struct lwan *l, const struct lwan_config *config)
 
     try_setup_from_config(l, config);
 
+    if (!lwan_strbuf_get_length(&l->headers)) {
+        lwan_headers_init(l, config->global_headers);
+    }
+
     lwan_response_init(l);
 
     /* Continue initialization as normal. */
@@ -597,6 +688,7 @@ void lwan_shutdown(struct lwan *l)
     lwan_status_debug("Shutting down URL handlers");
     lwan_trie_destroy(&l->url_map_trie);
 
+    lwan_strbuf_free(&l->headers);
     free(l->conns);
 
     lwan_response_shutdown(l);
--- a/src/lib/lwan.h
+++ b/src/lib/lwan.h
@@ -377,6 +377,7 @@ struct lwan_request {
     struct lwan_value url;
     struct lwan_value original_url;
     struct lwan_connection *conn;
+    const struct lwan_strbuf *const global_response_headers;
     struct lwan_proxy *proxy;
 
     struct timeout timeout;
@@ -457,6 +458,7 @@ struct lwan_straitjacket {
 struct lwan_config {
     /* Field will be overridden during initialization. */
     enum lwan_request_flags request_flags;
+    struct lwan_key_value *global_headers;
 
     char *listener;
     char *error_template;
@@ -478,6 +480,7 @@ struct lwan_config {
 struct lwan {
     struct lwan_trie url_map_trie;
     struct lwan_connection *conns;
+    struct lwan_strbuf headers;
 
     struct {
         pthread_barrier_t barrier;
--- a/src/bin/testrunner/testrunner.conf
+++ b/src/bin/testrunner/testrunner.conf
@@ -125,3 +125,7 @@ listener *:8080 {
             cache for = ${CACHE_FOR:5}
     }
 }
+headers {
+    server = lwan/testrunner
+    x-global-header = present
+}
--- a/src/lib/lwan-response.c
+++ b/src/lib/lwan-response.c
@@ -354,9 +354,12 @@ size_t lwan_prepare_response_header_full(
             "\r\nAccess-Control-Allow-Headers: Origin, Accept, Content-Type");
     }
 
-    APPEND_CONSTANT("\r\nServer: lwan\r\n\r\n\0");
+    assert(request->global_response_headers);
+    assert(request->global_response_headers->buffer);
+    assert(request->global_response_headers->used);
+    APPEND_STRING_LEN(request->global_response_headers->buffer, request->global_response_headers->used);
 
-    return (size_t)(p_headers - headers - 1);
+    return (size_t)(p_headers - headers);
 }
 
 #undef APPEND_CHAR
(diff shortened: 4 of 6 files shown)


src/lib/lwan.c
```diff
@@ -597,6 +654,23 @@ void lwan_shutdown(struct lwan *l)
...
 
+    if (UNLIKELY(!!l->config.global_headers)) {
+        struct lwan_key_value *kv = l->config.global_headers;
+        struct lwan_key_value *s;
+        for (s = kv; s && s->key; ++s);
+        while (kv) {
+            struct lwan_key_value *kvp = kv++;
+            if (kvp >= s) {
+                break;
+            }
```
lpereira, line 665: Why are you not using `LWAN_ARRAY_FOREACH`?

src/lib/lwan.h
```diff
@@ -377,6 +377,7 @@ struct lwan_request {
     struct lwan_value url;
     struct lwan_value original_url;
     struct lwan_connection *conn;
+    struct lwan_strbuf *headers;
```
lpereira, line 380: Please mark this one as `const`:

```suggestion
    const struct lwan_strbuf *const headers;
```

So that it's not possible to change it.

Also, would be nice if it could be renamed to something else; `headers` can be easily confused with the request headers from the client. Maybe `response_headers` or something like that?

src/lib/lwan.c
```diff
@@ -597,6 +654,23 @@ void lwan_shutdown(struct lwan *l)
     lwan_status_debug("Shutting down URL handlers");
     lwan_trie_destroy(&l->url_map_trie);
 
+    if (UNLIKELY(!!l->config.global_headers)) {
```
lpereira, line 657: No need to use `!!`, or even `UNLIKELY()` here.  This isn't a fast path (happens once during initialization).  It's harmless, but not necessary either.

src/lib/lwan.c
```diff
@@ -543,6 +582,24 @@ void lwan_init_with_config(struct lwan *l, const struct lwan_config *config)
...
+                lwan_strbuf_append_printf(l->headers, "\r\n%s: %s", kv->key, kv->value);
+            }
+        }
+    }
+
+    if (LIKELY(!set_server)) {
+        lwan_strbuf_append_strz(l->headers, "\r\nServer: lwan");
+    }
+
+    lwan_strbuf_append_strz(l->headers, "\r\n\r\n");
```
lpereira, line 601: Please move this to a separate function.

src/lib/lwan.c
```diff
@@ -416,6 +452,9 @@ static bool setup_from_config(struct lwan *lwan, const char *path)
                 }
             } else if (streq(line->key, "straitjacket")) {
                 lwan_straitjacket_enforce_from_config(conf);
+            } else if (streq(line->key, "headers")) {
+                struct lwan_key_value_array headers;
```
lpereira, line 456: Using a `struct lwan_key_value_array` is an implementation detail of `parse_global_headers()`, so you can declare it there instead.

src/lib/lwan.c
```diff
@@ -132,6 +132,42 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
...
+        case CONFIG_LINE_TYPE_SECTION_END:
+            kv = lwan_key_value_array_append(hdrs);
+            kv->key = NULL;
+            kv->value = NULL;
+
+            lwan->config.global_headers = lwan_key_value_array_get_array(hdrs);
+            return;
+        }
+    }
+
```
lpereira, line 166:
```suggestion
`config_error(c, "EOF while looking for end of 'headers' section")`
```

src/lib/lwan.c
```diff
@@ -132,6 +132,42 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
...
+error:
+    lwan_key_value_array_reset(hdrs);
```
lpereira, line 168: While you'll only get here on error, and the program will end when that happens, might be a good idea to `LWAN_ARRAY_FOREACH()` and `free()` each key/value pair.

src/lib/lwan.c
```diff
@@ -132,6 +132,42 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
...
+    lwan_key_value_array_init(hdrs);
+
+    while ((l = config_read_line(c))) {
+        switch (l->type) {
+        case CONFIG_LINE_TYPE_SECTION:
+            config_error(c, "No sections are supported under the headers section");
+            goto error;
+
+        case CONFIG_LINE_TYPE_LINE:
+            kv = lwan_key_value_array_append(hdrs);
```
lpereira, line 152: `lwan_key_value_array_append()` can return NULL if allocation fails.

```suggestion
            kv = lwan_key_value_array_append(hdrs);
            if (!kv)
            	lwan_status_critical_perror("Could not allocate memory for custom response header");
```

src/lib/lwan.c
```diff
@@ -132,6 +132,42 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
...
+            kv->key = strdup(l->key);
+            kv->value = strdup(l->value);
```
lpereira, line 154: Same thing for `strdup()`: they may return NULL.

If you call `lwan_status_critical_perror()`, the program will end, so you can just do something like `if (!kv->key) critical(...)` and it should be fine.  (This is different to configuration errors, where cleanup still must be performed, because how configuration errors are handled are different.)

src/lib/lwan.c
```diff
@@ -543,6 +582,24 @@ void lwan_init_with_config(struct lwan *l, const struct lwan_config *config)
 
     try_setup_from_config(l, config);
 
+    l->headers = lwan_strbuf_new();
+    bool set_server = false;
+    if (UNLIKELY(!!l->config.global_headers)) {
+        for (struct lwan_key_value *kv = l->config.global_headers; kv->key; ++kv) {
+            set_server = !strcasecmp(kv->key, "server");
```
lpereira, line 589: Please use something like the following, as the code as written won't work unless `Server`  is the last key in the array.

```suggestion
            if (!strcasecmp(kv->key, "server"))
            	set_server = true;
```

src/lib/lwan.c
```diff
@@ -543,6 +582,24 @@ void lwan_init_with_config(struct lwan *l, const struct lwan_config *config)
 
     try_setup_from_config(l, config);
 
+    l->headers = lwan_strbuf_new();
+    bool set_server = false;
+    if (UNLIKELY(!!l->config.global_headers)) {
```
lpereira, line 587: No need to use `!!` or `UNLIKELY()` here.

src/lib/lwan.c
HalosGhost: Because, by this point in the program, the lwan-array is actually gone (the base of the array is casted and assigned to the config's ``struct lwan_key_value *``, but the array struct itself is out-of-scope). You're right in that this is doing effectively what `LWAN_ARRAY_FOREACH` is meant to do, I just couldn't think of a reasonable way of preserving the array struct till shutdown.

lpereira: Ah, you're right.  You can do something like `for (struct lwan_key_value *kv = l->config.global_headers; kv->key; kv++) { free(kv->key); free(kv->value); }`, and only then free up the `custom_headers` array.  Freeing `kvp` on each iteration isn't valid code (I recomment building the debug version with address sanitizer.)

lpereira: Also, what happens when someone isn't reading from the configuration file, but provided this in the config struct?  This is going to free memory that has been potentially allocated without malloc.

HalosGhost: mm, good point. I'll need to think on that for a bit.

HalosGhost: for documentation's sake, the solution to this problem is to have `lwan_headers_init()` take a second argument (a `struct lwan_key_value[]`); then, it is up to the caller to handle allocation of the array. This means `lwan_shutdown()` doesn't need to worry about `free()`ing the array, and that the `struct lwan_key_value_array` (from the suggestion below) can be internalized to `parse_global_headers()` without issue.

src/lib/lwan.c
```diff
@@ -132,6 +132,83 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
...
 
+static void lwan_headers_init(struct lwan *l, struct lwan_key_value *hdrs)
+{
+    assert(l);
+
+    l->headers = lwan_strbuf_new();
+    bool set_server = false;
+    if (hdrs) {
+        for (struct lwan_key_value *kv = hdrs; kv->key; ++kv) {
+            set_server |= !strcasecmp(kv->key, "server");
```
lpereira, line 143: I'd prefer if this were just a `if (!strcasecmp(...)) set_server = true;` instead.

src/lib/lwan.c
```diff
@@ -132,6 +132,83 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
...
+            config_error(
+                c, "No sections are supported under the 'headers' section");
+            goto cleanup;
+
+        case CONFIG_LINE_TYPE_LINE:
+            if (!((kv = lwan_key_value_array_append(&hdrs)))) {
+                lwan_status_critical_perror(
+                    "Could not allocate memory for custom response header");
+            }
+            if (!((kv->key = strdup(l->key)))) {
```
lpereira, line 178: There's no need to `strdup()` keys and values here: while this function is running, the whole configuration file is loaded into memory, so you can just store the pointers.  This simplifies quite a bit the logic here.

src/lib/lwan.c
```diff
@@ -132,6 +132,83 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
...
+
+    while ((l = config_read_line(c))) {
+        switch (l->type) {
+        case CONFIG_LINE_TYPE_SECTION:
+            config_error(
+                c, "No sections are supported under the 'headers' section");
+            goto cleanup;
+
+        case CONFIG_LINE_TYPE_LINE:
+            if (!((kv = lwan_key_value_array_append(&hdrs)))) {
```
lpereira, line 174: I'd prefer if this were written as separate lines:

```C
kv = lwan_key_value_array_append(&hdrs);
if (!kv) {
    ...
}
```

src/lib/lwan.c
```diff
@@ -132,6 +132,83 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
...
+                    "Could not allocate memory for custom response header");
+            }
+            if (!((kv->value = strdup(l->value)))) {
+                lwan_status_critical_perror(
+                    "Could not allocate memory for custom response header");
+            }
+            break;
+
+        case CONFIG_LINE_TYPE_SECTION_END:
+            if (!((kv = lwan_key_value_array_append(&hdrs)))) {
```
lpereira, line 189: Same here: please make this into separate lines.

src/lib/lwan.c
```diff
@@ -132,6 +132,83 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
...
+        }
+    }
+
+    config_error(c, "EOF while looking for end of 'headers' section");
+
+cleanup:
+    LWAN_ARRAY_FOREACH (&hdrs, kv) {
+        free(kv->key);
+        free(kv->value);
+    }
```
lpereira, line 208: This can go away when the `strdup()` calls are removed.

src/lib/lwan.c
```diff
@@ -132,6 +132,83 @@ static struct lwan_url_map *add_url_map(struct lwan_trie *t, const char *prefix,
     return copy;
 }
 
+static void lwan_headers_init(struct lwan *l, struct lwan_key_value *hdrs)
+{
+    assert(l);
+
+    l->headers = lwan_strbuf_new();
```
lpereira, line 139: Might be a good idea to embed the strbuf in the lwan struct instead, and call `lwan_strbuf_init()` here.  (This will remove one pointer dereference when building the response headers.  Not a whole lot of difference, but it's easy to avoid it.)

It's safe to call `lwan_strbuf_free()` even if you don't call `lwan_strbuf_new()`: the strbuf knows if it has been allocated with `lwan_strbuf_new()` and won't call `free()` if `malloc()` wasn't used to allocate memory for the strbuf struct.

src/lib/lwan.c
HalosGhost: per off-line conversation, keeping `strdup()` because config reader would overwrite the pointers. However, will break these into separate lines as with other feedback.

src/lib/lwan.c
HalosGhost: Keeping these since `strdup()`s are being kept.

PR #282, opened 2020-06-16
Add struct casting to LWAN_ARRAY_STATIC_INIT

asymness:
Without the struct castings, the `make` build is failing with the following error:
image


lpereira:
Thanks for the PR, but I already reverted the patch adding that `#define`.

asymness:
Problem solved. Thanks!

PR #296, opened 2020-11-14
docs: fix simple typo, wich -> which

timgates42:
There is a small typo in src/lib/list.h.

Should read `which` rather than `wich`.


lpereira:
Merged, thanks!

PR #300, opened 2021-01-04
Add application/wasm Web Assembly mime type

daboross:
This adds the application/wasm mime type, which is used to represent compiled Web Assembly files, and registers it with the associated file extension, `.wasm`.

It's not a registered standard mimetype, but it is widely recognized. It's the mimetype listed in the [official WebAssembly conventions document](https://webassembly.github.io/spec/core/binary/conventions.html), and all major browsers recognize it.

My main motivation for this is so that I can use the `WebAssembly.instantiateStreaming` browser API on wasm files served by lwan. This api specifically requires the application/wasm mimetype, otherwise the browser won't instantiate streaming and instead requires loading the wasm file as a binary value before initializing. See the MDN documentation mentioning this requirement: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiateStreaming#Instantiating_streaming

Hopefully this will be an official mimetype at some point - there's an issue for that here: https://github.com/WebAssembly/spec/issues/573.

However, until then - I saw this file listed both "official" and "widely recognized" mimetypes, and I'm hoping application/wasm falls into the second one.

### DIFF
--- a/src/bin/tools/mime.types
+++ b/src/bin/tools/mime.types
@@ -1000,6 +1000,7 @@ application/vnd.zul				zir zirz
 application/vnd.zzazz.deck+xml			zaz
 application/voicexml+xml			vxml
 # application/vq-rtcpxr
+application/wasm				wasm
 # application/watcherinfo+xml
 # application/whoispp-query
 # application/whoispp-response


lpereira:
LGTM.  Quite curious on what you're working on with Lwan and WASM; is it something public?

daboross:
Thanks!

I'm using lwan as a local deployment server for wasm game development. I figure it isn't exactly the intended use case, but it's A) quick to configure, B) an actually fully featured web server, and C) nicely efficient. I don't have anything public yet, but it's been really nice to use for local web development & prototyping out rust-wasm-based games.

FabioMen10:
I thank you @daboross for your explanation and information. @lpereira I'm learning to make the same use that @daboross has been doing. I and many others, I believe, need wasm support at LWAN. I may be getting it wrong and LWAN already has that.

lpereira:
@daboross Awesome! Good to know it's working for you!

@FabioMen10 Sort of adding the MIME Type in Lwan, like @daboross did, there's nothing much I can think of that could use WASM in Lwan (well, maybe some sort of CGI scripts and whatnot, but that's implementable using what's already there, and won't need changes in Lwan itself; considering there's not much of a standard here, I don't see the point at this time in implementing this).

PR #303, opened 2021-04-08
Adding PUT handler

pontscho:
Hey, I tried to add PUT support for lwan, I'm not hundred percent sure about that I was able to follow your thoughts in any ways, but this code is able to handle PUT requests both from browsers and curl without any problem. And I've also added some helper functions to the lua api, those new methods made my life easier. I would really appreciate your comments and thoughts about this patch. Thanks!

### DIFF
--- a/src/lib/lwan-lua.c
+++ b/src/lib/lwan-lua.c
@@ -143,8 +143,8 @@ LWAN_LUA_METHOD(cookie)
 LWAN_LUA_METHOD(body)
 {
     struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
-    if (request->helper->put_data.len) {
-        lua_pushlstring(L, request->helper->put_data.value, request->helper->put_data.len);
+    if (request->helper->body_data.len) {
+        lua_pushlstring(L, request->helper->body_data.value, request->helper->body_data.len);
     } else {
         lua_pushlstring(L, "", 0);
     }
--- a/src/lib/lwan-lua.c
+++ b/src/lib/lwan-lua.c
@@ -93,9 +93,53 @@ static int request_param_getter(lua_State *L,
     return 1;
 }
 
+LWAN_LUA_METHOD(remote)
+{
+	struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    char ip_buffer[INET6_ADDRSTRLEN];
+	lua_pushstring(L, lwan_request_get_remote_address(request, ip_buffer));
+	return 1;
+}
+
+
 LWAN_LUA_METHOD(header)
 {
     return request_param_getter(L, lwan_request_get_header);
+}
+
+LWAN_LUA_METHOD(method)
+{
+#define GEN_TABLE_ENTRY(upper, lower, mask, constant) [REQUEST_METHOD_##upper] = { .value = #upper, .len = sizeof(#upper) - 1 },
+    static const struct lwan_value method2name[REQUEST_METHOD_MASK] = {
+        FOR_EACH_REQUEST_METHOD(GEN_TABLE_ENTRY)
+    };
+#undef GEN_TABLE_ENTRY
+
+	struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+	lua_pushstring(L, method2name[lwan_request_get_method(request)].value);
+	return 1;
+}
+
+LWAN_LUA_METHOD(path)
+{
+	struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    if (request->url.value && request->url.len) {
+        lua_pushlstring(L, request->url.value, request->url.len);
+    } else {
+        lua_pushnil(L);
+    }
+	return 1;
+}
+
+LWAN_LUA_METHOD(query)
+{
+	struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    if (request->helper->query_string.value && request->helper->query_string.len) {
+        lua_pushlstring(L, request->helper->query_string.value, request->helper->query_string.len);
+    } else {
+        lua_pushnil(L);
+    }
+	return 1;
 }
 
 LWAN_LUA_METHOD(query_param)
@@ -111,6 +155,17 @@ LWAN_LUA_METHOD(post_param)
 LWAN_LUA_METHOD(cookie)
 {
     return request_param_getter(L, lwan_request_get_cookie);
+}
+
+LWAN_LUA_METHOD(body)
+{
+	struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    if (request->helper->put_data.value && request->helper->put_data.len) {
+        lua_pushlstring(L, request->helper->put_data.value, request->helper->put_data.len);
+    } else {
+        lua_pushnil(L);
+    }
+	return 1;
 }
 
 LWAN_LUA_METHOD(ws_upgrade)
--- a/src/lib/lwan-lua.c
+++ b/src/lib/lwan-lua.c
@@ -95,10 +95,10 @@ static int request_param_getter(lua_State *L,
 
 LWAN_LUA_METHOD(remote)
 {
-	struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
     char ip_buffer[INET6_ADDRSTRLEN];
-	lua_pushstring(L, lwan_request_get_remote_address(request, ip_buffer));
-	return 1;
+    lua_pushstring(L, lwan_request_get_remote_address(request, ip_buffer));
+    return 1;
 }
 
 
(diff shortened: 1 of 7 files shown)


lpereira:
Thanks for the PR, @pontscho! From a quick glance, the code looks good, but I'd appreciate if some refactoring were done before this gets merged.

src/lib/lwan-lua.c
```diff
@@ -93,11 +93,55 @@ static int request_param_getter(lua_State *L,
     return 1;
 }
 
+LWAN_LUA_METHOD(remote)
```
lpereira, line 96: Maybe call it `remote_address`?

src/lib/lwan-lua.c
```diff
@@ -93,11 +93,55 @@ static int request_param_getter(lua_State *L,
...
+    return 1;
+}
+
+
 LWAN_LUA_METHOD(header)
 {
     return request_param_getter(L, lwan_request_get_header);
 }
 
+LWAN_LUA_METHOD(method)
```
lpereira, line 110: While this is fine, what's the purpose of such method? Handlers are supposed to be already per method, so this seems redundant.

src/lib/lwan-lua.c
```diff
@@ -93,11 +93,55 @@ static int request_param_getter(lua_State *L,
...
+    struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    if (request->url.value && request->url.len) {
+        lua_pushlstring(L, request->url.value, request->url.len);
+    } else {
+        lua_pushnil(L);
+    }
+    return 1;
+}
+
+LWAN_LUA_METHOD(query)
```
lpereira, line 134:
```suggestion
LWAN_LUA_METHOD(query_string)
```

src/lib/lwan-lua.c
```diff
@@ -93,11 +93,55 @@ static int request_param_getter(lua_State *L,
...
+
+    struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    lua_pushstring(L, method2name[lwan_request_get_method(request)].value);
+    return 1;
+}
+
+LWAN_LUA_METHOD(path)
+{
+    struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    if (request->url.value && request->url.len) {
```
lpereira, line 126: `request->url.value` should never be NULL, so this method returning `nil` is kind of unexpected, I think. Maybe get rid of the `if`?

src/lib/lwan-lua.c
pontscho: Of course.

src/lib/lwan-lua.c
```diff
@@ -93,11 +93,55 @@ static int request_param_getter(lua_State *L,
...
+    } else {
+        lua_pushnil(L);
+    }
+    return 1;
+}
+
+LWAN_LUA_METHOD(query)
+{
+    struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    if (request->helper->query_string.value && request->helper->query_string.len) {
```
lpereira, line 137: Here, the query string can be NULL, so we can do two thing in this function; first, no need to check the value and length, just the length:

```suggestion
    if (request->helper->query_string.len) {
```

src/lib/lwan-lua.c
pontscho: I rewrote the router, there was this method useful. I will remove it.

src/lib/lwan-lua.c
```diff
@@ -93,11 +93,55 @@ static int request_param_getter(lua_State *L,
...
+        lua_pushlstring(L, request->helper->query_string.value, request->helper->query_string.len);
+    } else {
+        lua_pushnil(L);
```
lpereira, line 140: And push an empty string instead, so people don't have to deal with a surprising `nil`:

```suggestion
        lua_pushlstring(L, "", 0);
```

src/lib/lwan-lua.c
pontscho: Yes, I wasn't sure about that the url never can be NULL.

src/lib/lwan-lua.c
```diff
@@ -113,6 +157,17 @@ LWAN_LUA_METHOD(cookie)
     return request_param_getter(L, lwan_request_get_cookie);
 }
 
+LWAN_LUA_METHOD(body)
+{
+    struct lwan_request *request = lwan_lua_get_request_from_userdata(L);
+    if (request->helper->put_data.value && request->helper->put_data.len) {
```
lpereira, line 163: Same thing here: just check for `len` and instead of pushing `nil`, push an empty string.

src/lib/lwan-lua.c
pontscho: Okay

src/lib/lwan-lua.c
pontscho: Interesting, but okay :)

src/lib/lwan-private.h
```diff
@@ -36,8 +36,10 @@ struct lwan_request_parser_helper {
     struct lwan_value query_string;	/* Stuff after ? and before # */
 
     struct lwan_value post_data;	/* Request body for POST */
-    struct lwan_value content_type;	/* Content-Type: for POST */
+    struct lwan_value put_data;	/* Request body for PUT */
```
lpereira, line 39: This is the major thing I'd like to see changed in the patch: there's no difference between put and post data; maybe rename this to `body_data` and have only one?

src/lib/lwan-lua.c
pontscho: Roger.

src/lib/lwan-private.h
```diff
@@ -36,8 +36,10 @@ struct lwan_request_parser_helper {
...
+    struct lwan_value content_type;	/* Content-Type: for POST and PUT */
     struct lwan_value content_length;	/* Content-Length: */
+    struct lwan_value expect; /* Expect: */
```
lpereira, line 42: `Expect` is an uncommon header. Do we need a fast path here, or can we use the function to look up headers by name when we need it?

src/lib/lwan-request.c
```diff
@@ -1105,13 +1111,13 @@ free_post_buffer(void *data)
 }
 
 static void*
-alloc_post_buffer(struct coro *coro, size_t size, bool allow_file)
+alloc_buffer(struct coro *coro, size_t size, size_t thresh, bool allow_file)
```
lpereira, line 1114:
```suggestion
alloc_body_buffer(struct coro *coro, size_t size, size_t thresh, bool allow_file)
```

src/lib/lwan-request.c
```diff
@@ -1105,13 +1111,13 @@ free_post_buffer(void *data)
...
 {
     struct file_backed_buffer *buf;
     void *ptr = (void *)MAP_FAILED;
     int fd;
 
-    if (LIKELY(size < post_buffer_temp_file_thresh)) {
+    if (LIKELY(size < thresh)) {
```
lpereira, line 1120:
```suggestion
    if (LIKELY(size < body_buffer_temp_file_thresh)) {
```

src/lib/lwan-private.h
pontscho: That was my first thought too, but I wanted to keep separated the post and the put code path. I will change that, but I think not today. :)

src/lib/lwan-request.c
```diff
@@ -1250,6 +1259,55 @@ static enum lwan_http_status discard_post_data(struct lwan_request *request)
...
+    helper->error_when_n_packets = lwan_calculate_n_packets(total);
+
+    struct lwan_value buffer = {.value = new_buffer, .len = total};
+    return client_read(request, &buffer, total, post_data_finalizer);
+}
+
+static enum lwan_http_status discard_put_data(struct lwan_request *request)
+{
+    return discard_post_data(request);
+}
```
lpereira, line 1309: These are the same for the post path, so if we fold both in a single "body" concept, we can use one function for both.

src/lib/lwan-request.c
```diff
@@ -1395,6 +1453,14 @@ static enum lwan_http_status prepare_for_response(struct lwan_url_map *url_map,
...
     }
 
+    if (lwan_request_get_method(request) == REQUEST_METHOD_PUT) {
+        if (url_map->flags & HANDLER_HAS_PUT_DATA)
+            return read_put_data(request);
+
+        enum lwan_http_status status = discard_put_data(request);
+        return (status == HTTP_OK) ? HTTP_NOT_ALLOWED : status;
+    }
+
```
lpereira, line 1463: Maybe get rid of this if and do something like:

```c
switch (lwan_request_get_method(request)) {
case REQUEST_METHOD_PUT:
case REQUEST_METHOD_POST:
       if (url_map->flags & HANDLER_HAS_BODY_DATA)
              return read_body_data(request);
      ...
}
```

src/lib/lwan-private.h
pontscho: It has own rfc and the cURL is expecting that. This type of PUT is able to work without this header but in that case it has a timeout and I do not want to wait it.

src/lib/lwan-request.c
lpereira: Eventually we can dedicate a bit in one of the request method values to say that it can have body data, or something like this. Maybe only look at the `Content-Length` header, or something like this. No need to do any of this now, though.

src/lib/lwan-lua.c
pontscho: Okay.

src/lib/lwan-request.c
pontscho: I agree with that.

pontscho:
Thank you very much ! I've finished with the smaller changes, I hope tomorrow I will finish with the rest.

src/lib/lwan-private.h
lpereira: Right -- but you can use `lwan_request_get_header(request, "Expect")` for this, as it's not that common for all other requests.  This is already used for things like Cookies and whatnot.

pontscho: Wow, I will check it.

src/lib/lwan-request.c
pontscho: Renamed.

src/lib/lwan-request.c
pontscho: Renamed.

src/lib/lwan-request.c
pontscho: I made a common handler function but I kept the POST and the PUT separated because of its configuration options. I think this way makes lwan more flexible without any performance degradation and its fit for my personal goals more better.

lpereira:
From a cursory look, everything looks good. I'll take a closer look during the weekend before merging. Thanks for the patch!

Also, if you don't mind me asking, what are you building with Lwan? I'm always curious to see what people are building with it. If it's public, feel free to send a PR to the README file adding a link to it.

pontscho:
No, I thank you your work on lwan, this server was the first that was able integrating scripting and serving on that way I prefer.

I want to use it for "basement" of a site. I think law is powerful enough for this. I miss the TLS support, but varnish maybe be good enough for terminating the TLS connections, I will check it later. If not... I will consider implementing it in lwan beside a logging.

I really excited about your final answer. :)

lpereira:
Varnish doesn't do TLS termination, but you can use Hitch instead. It's developed by the same team that develops Varnish. I use it on the same server that runs the Lwan website, my blog, and a few other things, and it works pretty well.

I'm squashing some of your commits and splitting up the Lua changes from the changes adding PUT support, but otherwise, I'm merging your changes.

lpereira:
Merged. Thanks!

PR #315, opened 2021-09-12
timeout: Code cleanup

jserv:
This patch makes timeout library more consistent with Lwan codebase
including code style and header inclusion.


lpereira:
Thanks for the PR.  However, I'd rather keep this as close as possible to the original source code, so that if there's an update upstream, it's easier to just copy/paste whatever was done.

lpereira:
Closing the PR for the stated reason.

PR #328, opened 2022-01-21
Things

pontscho:
Hey,

I've made some small patches that makes my life easier, I thought I share with these with You. Maybe it can be useful for others too.

 * fixing build on macOS: I couldn't build lean on my notebook without these small changes.
 * enabling luajit 2.1: I've started to use openresty's fork about luajit, this version has a lot's of performance optims and they are maintaining it actively 
 * making brotli/zstd optional (enabled by default): zstd and brotli is installed on my notebook but doesn't really need these support for me, I've added two cmake options for it
 * allow remote address override by X-Real-IP: I've started to use nginx for TLS and HTTP2 termination, nginx supports more flexible routing and acl management supports than hitch but by default the remote address will be determined to a localhost. With adding X-Real-IP header possible to log out the real IP address.

Thanks,
pontscho


lpereira:
As far as the `X-Real-IP` header goes, it [seems like you can use the PROXY protocol on nginx for outgoing connections too](https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/#proxy-protocol-for-a-tcp-connection-to-an-upstream); Lwan already supports that.  I would prefer Lwan to support only one scheme for this feature.

lpereira:
Thanks for the PR! I'll look at them more carefully whenever I have some time.

pontscho:
> As far as the `X-Real-IP` header goes, it [seems like you can use the PROXY protocol on nginx for outgoing connections too](https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/#proxy-protocol-for-a-tcp-connection-to-an-upstream); Lwan already supports that. I would prefer Lwan to support only one scheme for this feature.

Yep, It's right, I've tried it, stream proxy is almost a good way but in that case I lost the HTTP2 termination and that is important for me.

src/lib/missing/sys/epoll.h
```diff
@@ -32,9 +32,9 @@ enum {
     EPOLLHUP = EPOLLRDHUP
 } epoll_event_flag;
 
-enum { EPOLL_CTL_ADD, EPOLL_CTL_MOD, EPOLL_CTL_DEL } epoll_op;
+typedef enum { EPOLL_CTL_ADD, EPOLL_CTL_MOD, EPOLL_CTL_DEL } epoll_op;
```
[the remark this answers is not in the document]
pontscho: Okay, I will check it.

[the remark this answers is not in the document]
pontscho: Yep, it okay.

lpereira:
Thanks for confirming! Will cherry pick the others, then

pontscho:
Hi,

I've added some other patches to my fork,

1. I increased the HTTP header buffer four times because 512 bytes wasn't enough in some cases in practice (for example I use a littlebit more complex cookie, content encoding and some other headers for debugging)
2. I've fixed epoll.h on macos
3. maybe this patch that you won't like :) I moved the read_cpu_topology() function to lwan.c and read this information from /sys before lwan call chroot(). With this change is possible to avoid needs for /sys in chrooted environment.

Thank you very much for other applying my patches. 
pontscho

lpereira:
I thank you for working on Lwan! I'll cherry pick these patches as soon as I have some free time

lpereira:
I would prefer if you sent a new PR for each kind pack of related changes you're making. If you keep adding commits to this PR, it'll be a bit hard for me to review. (I don't have much time to work on Lwan anymore, unfortunately.)

src/lib/lwan.c
```diff
@@ -773,6 +782,98 @@ static void get_number_of_cpus(struct lwan *l)
...
+
+    l->have_cpu_topology = true;
+}
+
+#elif defined(__APPLE__)
+static void read_cpu_topology(struct lwan *l)
+{
+    size_t length = sizeof(int);
+    int physical_cores = 1; sysctlbyname("hw.physicalcpu", &physical_cores, &length, NULL, 0);
+    int logical_cores = 1; sysctlbyname("hw.logicalcpu", &logical_cores, &length, NULL, 0);
```
lpereira, line 861: There's no need to write code that is this terse; please spread this out and do error checking when calling `sysctlbyname()`.

src/lib/lwan.c
```diff
@@ -773,6 +782,98 @@ static void get_number_of_cpus(struct lwan *l)
...
+    bool ht_enabled = physical_cores != logical_cores;
+
+    l->have_cpu_topology = false;
+
+    for (int i = 0;i < logical_cores;i++) {
+        l->cpu_siblings[i] = i / ((int)ht_enabled + 1);
+    }
```
lpereira, line 868: This won't work in all cases -- it's not always that `cpu[i]` is a sibling of `cpu[i+1]`.  CPU topologies vary quite a bit; that's why it's that complex to do it on the Linux version of this code. (I also suspect that the Linux code is a bit wrong here; it doesn't seem to give the correct results on my laptop.  I don't have access to -- or care about -- NUMA machines and that may make things even more wrong.)

The topology is important in the scheduling code because of the way memory is allocated and the connection array is laid out; it's made like this to avoid [false sharing](https://en.wikipedia.org/wiki/False_sharing) when accessing that array. (There are other reasons, including the usage of `SO_INCOMING_CPU` or the CBPF code that hints the kernel to process the interrupts in the same CPU that will handle the connection in Lwan to avoid having to potentially migrate stuff between cores.)

src/lib/lwan-websocket.c
```diff
@@ -56,43 +56,41 @@ enum ws_opcode {
...
-        frame[0] = (uint8_t)len;
-        frame_len = 1;
+        frame[1] = (uint8_t)len;
+        frame_len = 2;
     } else if (len <= 65535) {
-        frame[0] = 0x7e;
-        memcpy(frame + 1, &(uint16_t){htons((uint16_t)len)}, sizeof(uint16_t));
-        frame_len = 3;
+        frame[1] = 0x7e;
+        *((uint16_t *)(frame + 2)) = htons((uint16_t)len);
```
lpereira, line 71: This can be an issue on systems where unaligned access is either not possible, or there's a performance penalty for doing so. It's better to just use `memcpy()`; the compiler knows about it and can generate similar code if the length is known at compile time.

src/lib/lwan-thread.c
```diff
@@ -1191,16 +1140,13 @@ void lwan_thread_init(struct lwan *l)
     uint32_t n_threads;
     bool adj_affinity;
 
-#if defined(__x86_64__) && defined(__linux__)
+#if (defined(__x86_64__) && defined(__linux__)) || defined(__APPLE__)
```
lpereira, line 1143: This `#ifdef` thing is starting to look like a maze. Might be easier to move this to a file of its own where it's clearer what's happening.

src/lib/lwan-request.c
```diff
@@ -1650,6 +1653,10 @@ lwan_request_get_remote_address(struct lwan_request *request,
 
             return memcpy(buffer, unspecified, sizeof(unspecified));
         }
+    } else if (request->flags & REQUEST_ALLOW_REMOTE_ADDRESS_OVERRIDE && request->helper->xrealip.value && request->helper->xrealip.len) {
+        size_t len = LWAN_MIN(request->helper->xrealip.len, INET6_ADDRSTRLEN - 1);
+        strncpy(buffer, request->helper->xrealip.value, len), buffer[len] = 0;
```
lpereira, line 1658: Please write one statement per line. I missed the `buffer[len]` the first time I read this.

src/lib/lwan-request.c
```diff
@@ -575,6 +575,9 @@ static bool parse_headers(struct lwan_request_parser_helper *helper,
         case STR4_INT_L('H', 'o', 's', 't'):
             SET_HEADER_VALUE(host, "Host");
             break;
+        case STR4_INT_L('X', '-', 'R', 'e'):
+            SET_HEADER_VALUE(xrealip, "X-Real-IP");
+            break;
```
lpereira, line 580: This function is a really hot spot while parsing requests, and I'd like to really think about adding things here that most users won't need -- especially since Lwan also supports the PROXY protocol that covers this use case.

If this ends up being integrated, I'd rather see something like what we do for Cookies, where we look for the header if and only if a handler requests it, parse them, and sets a flag to avoid having to look for the header again. I'm not really convinced, however, that also supporting this header is necessary, either, given how prevalent PROXY protocol is.

src/lib/lwan.c
```diff
@@ -591,6 +596,9 @@ static bool setup_from_config(struct lwan *lwan, const char *path)
             } else if (streq(line->key, "proxy_protocol")) {
                 lwan->config.proxy_protocol =
                     parse_bool(line->value, default_config.proxy_protocol);
+            } else if (streq(line->key, "allow_remote_address_override")) {
+                lwan->config.allow_remote_address_override =
+                    parse_bool(line->value, default_config.allow_remote_address_override);
```
lpereira, line 601: Given we also support the PROXY protocol for the same thing, it might be a better idea to see a single configuration option for both features (e.g. `allow_spoofing_origin_ip = [x-real-ip] [PROXY]`).

src/lib/lwan-websocket.c
```diff
@@ -56,43 +56,41 @@ enum ws_opcode {
...
-        {.iov_base = &header_byte, .iov_len = 1},
         {.iov_base = frame, .iov_len = frame_len},
         {.iov_base = msg, .iov_len = len},
     };
 
     lwan_writev(request, vec, N_ELEMENTS(vec));
 }
 
-void lwan_response_websocket_write(struct lwan_request *request)
+void lwan_response_websocket_write(struct lwan_request *request, bool is_text)
```
lpereira, line 87: For legibility reasons, it might be better to have two new functions: `lwan_response_websocket_write_text()`and `lwan_response_websocket_write_binary()`, calling this function with either a `WS_OPCODE_TEXT` or `WS_OPCODE_BINARY` parameter. This avoids the branch and removes a boolean parameter to a function (which is sometimes tricky to reason about in the calling site).

lpereira (changes requested):
Thanks for the PR! I appreciate you hacking on this, especially while trying to improve macOS support -- which I never use, so it's frequently broken, especially now that the macOS build bot is not active.

I do have some questions and comments, however:

src/lib/lwan-websocket.c
pontscho: First time I created same API both for C and Lua API (ws_text(), ws_binary(), etc) but I wanted keep backward compatibility with current existing API live ws_write. I will change it back for my original thoughts.

src/lib/lwan.c
pontscho: Okay.

src/lib/lwan-request.c
pontscho: Okay.

src/lib/lwan.c
pontscho: In practice those function calls never fails but I will do.

src/lib/lwan.c
pontscho: I will make this change if you keep this x-real-ip header change. :)

PR #344, opened 2022-05-02
WebSocket: handling masked ping on proper way

pontscho:
Hi,

in some cases, for example uses k6 (https://k6.io) for performance testing, the client on the other side sends masked ping message and it uses strict protocol checking, in that case Lwan has to send back payload of ping in a pong message unmasked and has to set the FIN flag on the header. In other cases the other side will drop the connection immediately.

I've fixed this issue based on the original behaviour of Lwan.

pontscho


lpereira:
Patch looks good to me! I'll merge it when I'm near my laptop again. Thanks!

PR #358, opened 2023-07-19
Add Podman socket activation example

eriksjolund:
Add Podman socket activation example in _README.md_

I'm not exactly sure about the text "_Requirements: Podman version 4.5.0 or higher._". (I used that version for testing)


lpereira:
Merged a long time ago.

PR #365, opened 2024-01-31
Define MSG_FASTOPEN

yonas:
For some reason, this gets picked up on FreeBSD and not Linux.

Also, I'm not 100% sure `MSG_FASTOPEN` should be defined as `0` by default.


lpereira:
Merged, thanks.  This is related to TCP Fastopen, so should be fine if that mechanism isn't being used on FreeBSD.

PR #378, opened 2026-05-25
Fix condition check for character range validation

### LINKED ISSUE #377: config: `_parse_i64()` accepts invalid first bytes
I found this while fuzzing the configuration parser, then confirmed it in the normal config parsing path with Valgrind and source review.

After skipping an optional leading `-`, `_parse_i64()` checks the first remaining byte like this:

```c
if (UNLIKELY(*s < '0' && *s > '9'))
    return false;
```

The condition is impossible, because one byte cannot be both below `'0'` and above `'9'`. As a result, `_parse_i64()` accepts any first byte, adds it to the accumulator as `*s - '0'`, and only validates the following bytes after that first byte has already been consumed.

This was introduced in `cb33c956` (`Use homegrown integer parsing utilities`, 2025-12-02 19:42:25 -0300), and is still present in current `master` at `901f5b0`.

## Reproducer

For example, this config gives `quiet` an empty value:

```text
quiet =
```

In normal config loading, `lwan_load_config()` parses `quiet` with `parse_bool()`. Exercising the same `config_open()` and `parse_bool()` path under Valgrind gives:

```text
Conditional jump or move depends on uninitialised value(s)
    at _parse_i64 (src/lib/lwan-config.c:173)
    by parse_bool (src/lib/lwan-config.c:247)

Uninitialised value was created by a heap allocation
    by grow_buffer_if_needed_internal (src/lib/lwan-strbuf.c:70)
```

The empty value starts with `NUL`. `_parse_i64()` consumes that byte as `'\0' - '0'`, advances `s`, and then reads the byte after the logical end of the string. That byte is still inside the allocated `lwan_strbuf` capacity, but it is not part of the current string.

## Invalid Numeric Values

The same first-byte check also accepts malformed numeric strings. I observed these values through the same parser helpers:

```text
f        -> 54
X959     -> 40959
/        -> -1
-/       -> 1
~9999    -> 789999
```

## Reachable Code Paths

The bug is reachable from Lwan config options that use `parse_bool()` or `parse_long()`. For example:

- `allow_cors = f` and `proxy_protocol = f` parse as true.
- `max_post_data_size = ~9999` parses as 789999, so a malformed config value can raise the configured POST body limit.

It is also reachable from HTTP request parsing when a handler uses `HANDLER_EXPECTS_BODY_DATA`:

```text
maybe_read_body_data()
  -> read_body_data()
  -> get_remaining_body_data_length()
  -> parse_long_long(helper->content_length.value, -1)
```

For example, `Content-Length: f` is accepted as a 54-byte body length.

## Impact

I looked for stronger impact from these primitives, including memory disclosure, reads past the allocation boundary, DoS against a default Lwan deployment, and request-smuggling style proxy chains. I did not find a practical exploit path inside Lwan itself. The remote `Content-Length` case is still bounded by the configured body-size limit, and the config examples require influence over the configuration file or a config-generation layer.

So I am reporting this as a numeric parser bug rather than as a security issue.

Carmel0:
Fixes #377 

This changes the check to `||`, so empty values and non-digit first bytes are rejected before the parser consumes them.

### DIFF
--- a/src/lib/lwan-config.c
+++ b/src/lib/lwan-config.c
@@ -165,7 +165,7 @@ static bool _parse_i64(const char *s, int64_t *out)
         negative = true;
     }
 
-    if (UNLIKELY(*s < '0' && *s > '9'))
+    if (UNLIKELY(*s < '0' || *s > '9'))
         return false;
 
     goto elide_mult_for_first_iter;


lpereira:
Thank you! I don't know how this one passed through. :)

